PR #46, opened 2023-10-17
Add Lee's code (CLI + Various Fixes) to develop.

LeeBergstrand:
Add CLI as well as various bug fixes.


LeeBergstrand:
@jmtsuji I would like to move my code into development.

README.md
```diff
@@ -1,60 +1,66 @@
...
+
 ### 1. Install
+
 ```bash
 git clone https://github.com/jmtsuji/rotary.git
-conda env create -n rotary --file=rotary/envs/rotary.yaml
+conda env create -n rotary --file=rotary/envs/enviroment.yaml
 ```
 
 ### 2. Copy and fill out the config (YAML) file (`config.yaml`).
```
jmtsuji, line 66: We'll need to update this section later -- I assume `rotary init` now generates the config file?
I think the README can be further updated in future before pushing to master and does not need to be fully updated before this PR is merged to develop.

README.md
```diff
@@ -63,20 +69,24 @@ The basic parameters to fill are the genome name, the paths to the long and (QC'
...
 
 Lastly, make sure you set the threads and memory to values that make sense for your server.
 
 One you're finished editing the config file, save it as something like `myconfig.yaml`.
 
-**Note**: short read QC is not performed by _rotary_, so you'll want to do short read QC prior to using this pipeline. 
+**Note**: short read QC is not performed by _rotary_, so you'll want to do short read QC prior to using this pipeline.
 I recomend using the qc module of the [ATLAS pipeline](https://github.com/metagenome-atlas/atlas) for quick/robust QC.
 
 ### 3. Run
```
jmtsuji, line 88: Similarly to above, we'll need to update this section in future.

enviroment.yaml
```diff
@@ -0,0 +1,13 @@
+channels:
+  - conda-forge
+  - bioconda
+  - defaults
+dependencies:
+  - python=3.9.*
+  - pandas=1.4.*
```
jmtsuji, line 7: We should probably bump the versions of these dependencies before pushing to master in future (e.g., pandas could now be bumped to pandas 2)

rotary/rotary.py
```diff
@@ -0,0 +1,328 @@
...
+                                         file_names=run_files)
+
+    if run_file_presences == [False, False]:
+        has_run_files = False
+    elif run_file_presences == [True, True]:
+        has_run_files = True
+    else:
+        raise ValueError('The output directory is missing a run configuration file: {}'.format(zip(run_files,
+                                                                                                   run_file_presences)))
+    existing_message = 'Existing run configuration files {} in the output directory.'.format(run_files)
```
jmtsuji, line 96: This isn't necessary to change, but are you aware of f-strings? They were introduced in Python 3.6 and are a more concise way to format text strings. See https://docs.python.org/3.9/tutorial/inputoutput.html#tut-f-strings and https://peps.python.org/pep-0498/
I recently discovered them and really like them for their readability.

rotary/sample.py
```diff
@@ -0,0 +1,93 @@
...
+class SequencingFile(object):
+    """
+    An object representing a FASTQ file.
+    """
+
+    def __init__(self, file_path):
+        self.path = file_path
+        self.name = os.path.basename(self.path)
+
+        if 'fastq' not in self.name:
```
jmtsuji, line 23: Perhaps `.fq` should also be OK?

rotary/sample.py
```diff
@@ -0,0 +1,93 @@
...
+            raise ValueError('{} is not a fastq file.'.format(self.name))
+
+        if '_' in self.name:
+            identifier, remaining = self.name.split('_', 1)  # Split on first _ if found.
```
jmtsuji, line 27: I can think of cases where keeping the text before the first underscore could lead to naming collisions between different samples (e.g., if someone had samples like `P_putida`, `P_aeruginosa`, etc.). Might it be better to strip off the final R1 / R2 and then keep everything before that as the sample ID? Not sure...

rotary/utils.py
```diff
@@ -0,0 +1,52 @@
...
+        cli_path = getattr(args, argument)
+        if cli_path:
+            cli_path = sanitize_cli_path(cli_path)
+    else:
+        cli_path = None
+
+    return cli_path
+
+
+def sanitize_cli_path(cli_path):
```
jmtsuji, line 43: This is smart!

jmtsuji (commented):
These are great additions to rotary -- thanks for building the CLI, working on the install, and adding bug fixes to the workflow. I've flagged some minor areas where I had comments/questions about the new code, but overall, this PR looks like it is in good shape. Let me know your thoughts -- thanks!

README.md
LeeBergstrand: I agree that the readme needs to be updated before the final release. Yes, `rotary init` is the as `atlas init`. It makes the `samples.tsv` and the `config.yaml` and places them in the project directory. You can point `rotary init` at a folder of fastq files.

rotary/rotary.py
LeeBergstrand: @jmtsuji Nice, I didn't know about this. I have modified most of them. Seems like a good way to make the code less wordy.

README.md
LeeBergstrand: Added as issue.

README.md
LeeBergstrand: Added as issue.

rotary/rotary.py
```diff
@@ -0,0 +1,328 @@
...
+                            help='path the output/rotary project directory')
+    parser_run.add_argument('-j', '--jobs', metavar='JOBS',
+                            help='number of threads that rotary should use (overrides config)')
+    parser_run.add_argument('-s', '--snakemake_args', metavar='',
+                            help="quoted string with arguments to be passed to snakemake. i.e., -s'--dag' (no space after -s)")
+    # Add attribute to tell main() what sub-command was called.
+    parser_run.set_defaults(run=True)
+
+    # =========================
+    # Declare Run_One Sub-command
```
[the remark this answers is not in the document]
LeeBergstrand: Added as issue

rotary/rules/rotary.smk
```diff
@@ -205,6 +205,31 @@ rule validate_gtdb:
...
+        "../envs/gtdbtk.yaml"
+    log:
+        "logs/download/gtdb_db_mash.log"
+    benchmark:
+        "benchmarks/download/gtdb_mash.txt"
+    threads:
+        config.get("threads",1)
+    params:
+        fast_ani_genomes_dir= os.path.join(config.get("db_dir"),"GTDB_" + VERSION_GTDB_COMPLETE, 'fastani', 'database'),
+        ref_genome_path_list=os.path.join(config.get("db_dir"),"GTDB_" + VERSION_GTDB_COMPLETE + '_mash', 'ref_mash_genomes.txt')
```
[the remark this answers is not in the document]
LeeBergstrand: @jmtsuji Thanks for the tips this has been addressed.

rotary/rules/rotary.smk
```diff
@@ -359,8 +384,8 @@ rule assembly_end_repair:
 #        In particular, the format of assembly_info.txt will need to be standardized (also in assembly_end_repair).
 rule finalize_assembly:
     input:
-        assembly="assembly/end_repair/repaired.fasta",
-        info="assembly/end_repair/assembly_info.txt"
+        assembly="assembly/flye/assembly.fasta",
```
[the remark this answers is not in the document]
LeeBergstrand: Created issue.

rotary/sample.py
LeeBergstrand: Addressed.

rotary/sample.py
LeeBergstrand: Moved to issue

enviroment.yaml
LeeBergstrand: Added as issue

rotary/rules/rotary.smk
```diff
@@ -129,21 +130,20 @@ rule download_eggnog_db:
...
     log:
         "logs/download/gtdb_db_download.log"
     benchmark:
         "benchmarks/download/gtdb_db_download.txt"
     params:
         db_dir_root=os.path.join(config.get("db_dir")),
-        initial_download_dir=os.path.join(config.get("db_dir"), "release" + VERSION_GTDB + "_v2"),
-        db_dir=os.path.join(config.get("db_dir"), "GTDB_" + VERSION_GTDB),
-        url="https://data.gtdb.ecogenomic.org/releases/release" + VERSION_GTDB + "/" + VERSION_GTDB + ".0/auxillary_files/gtdbtk_r" + VERSION_GTDB + "_v2_data.tar.gz"
+        initial_download_dir=os.path.join(config.get("db_dir"), "release" + VERSION_GTDB_MAIN),
```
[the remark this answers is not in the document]
LeeBergstrand: @jmtsuji I'm pretty sure when you unzip the tar file that is downloaded the name of the resulting directory is release214 even if the sub-variant of the release is 214.1. I'll check during end to end testing. Below we used mv to rename the release file and if the resulting folder after download and de-tar is named wrong it should error out.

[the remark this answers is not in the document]
jmtsuji: Ah, I think you are right about the dir name after download. Agreed, if there is an issue, it should appear during end-to-end testing.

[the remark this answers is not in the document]
jmtsuji: @LeeBergstrand FYI, the download of the GTDB passes in an end-to-end test I ran that includes this updated code.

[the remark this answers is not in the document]
LeeBergstrand: Tremendous

Issue #47, opened 2023-10-17
Modify python scripts so they can be called directly rather than downloaded.


LeeBergstrand:
### Problem Description

The Python scripts for end repair are installed by downloading them from GitHub. 

```
rule install_internal_scripts:
    output:
        end_repair=os.path.join(config.get("db_dir"), "rotary-" + VERSION, "scripts", "flye_end_repair.sh"),
        end_repair_utils=os.path.join(config.get("db_dir"), "rotary-" + VERSION, "scripts", "flye_end_repair_utils.py"),
        install_finished=os.path.join(config.get("db_dir"), "checkpoints", "internal_scripts_" + VERSION)
    log:
        "logs/download/install_internal_scripts.log"
    benchmark:
        "benchmarks/download/install_internal_scripts.txt"
    params:
        db_dir=config.get("db_dir"),
        url="https://github.com/jmtsuji/rotary/archive/refs/tags/" + VERSION + ".tar.gz"
    shell:
        """
        mkdir -p {params.db_dir}
        wget -O - {params.url} 2> {log} | tar -C {params.db_dir} -xzf - >> {log} 2>&1
        touch {output.install_finished}
        """
```

Because the scripts are already included in the repo, they should be installed during installation rather than downloaded from Github. 

### Problem Solution

Because all the scripts are being converted to Python, we can install them alongside `rotary.py` using our `pyproject.toml` file. However, it should be noted that the argsparse code must be placed in its function as I did in `rotary.py`. When installing via `pyproject.toml`, the resulting CLI scripts call functions in the Python files directly rather than the overall Python file.

```
[project.scripts]
rotary = "rotary.rotary:main"
```

See: https://github.com/jmtsuji/rotary/blob/99209d8158a6df68949fecd21180787562ab35b2/rotary/rotary.py#L34

LeeBergstrand:
@jmtsuji This was weird and confusing for me. So basically, when pip installs via `pyproject.toml`, it creates a script that calls the `main()` function in your Python code. Because of this, other overall Python script is not called, so you must include the args parsing code as a dedicated function called by `main().`

For an example, See: https://github.com/jmtsuji/rotary/blob/develop_lee/rotary/rotary.py

jmtsuji:
Yes, agreed, we should properly install `repair.py` rather than downloading it from Github. Thanks for the advice... I'll try to format the script like you mentioned, with a dedicated parser function called by main().

jmtsuji:
[...]

PR #48, opened 2023-10-18
End repair

jmtsuji:
Basic working version of `repair.py`. An end-to-end test (not included in this PR) works for the standard use case and if circularization is not successful. Some TODO's still remain for `repair.py`, but I am making this PR with the MVP of `repair.py` so that work on the rotary pipeline overall can continue (e.g., for the install).

Note that this PR does not include updated snakemake code either -- I can make this in a future PR (ran out of time for this today).

@LeeBergstrand Feel free to close this PR if you'd rather have a more complete version of `repair.py` before continuing (a few edge cases still need to be addressed, for example, and as mentioned above, the snakemake code is not updated). Otherwise, feel free to review this PR. Thanks!


jmtsuji:
Important step to address #47

LeeBergstrand (approved):

Issue #58, opened 2023-10-19
Add Unified End Repair Python Script


LeeBergstrand:
Unify, Bug Fix, and improve end repair by having it implemented as a python script. This work is currently happening on: https://github.com/jmtsuji/rotary/tree/end_repair

jmtsuji:
The updated `repair.py` (called from the command line as `rotary-repair`) is now working in the rotary pipeline! I ran an end-to-end test that seems to have run fine through the assembly section of the workflow (although see notes at the bottom).

I had to add four assembly-related tools to the [`environment.yaml`](https://github.com/jmtsuji/rotary/blob/end_repair/enviroment.yaml) file to be able to run `rotary-repair`. One of these, `circlator`, is no longer being actively maintained and has a huge number of dependencies that aren't needed by this project. My long-term hope is to replace `circlator` with in-house code as discussed in #26 -- this should simplify the rotary install substantially. However, the other three tools will need to remain in the environment.yaml.

I think that this end_repair branch is tentatively OK to merge into develop, so I will make a PR for you to review.

Note: When I ran the end-to-end test, I tested commit [9a4678b](https://github.com/jmtsuji/rotary/commit/9a4678b9a03865c2f295079feedc368139fce0a2), which is one commit before what is currently the latest commit, [fe7481f](https://github.com/jmtsuji/rotary/commit/fe7481f2ec0470eb7eb44f43944cc8bf028ee2ab). Commit 9a4678b did not yet have the rotary-repair dependencies as part of environment.yaml, and I instead manually installed rotary-repair in the conda env that was made for the end repair step. In fe7481f, I got rid of the conda env for the end repair step and moved the dependencies to environment.yaml. Because I had to head home at that point and didn't have access to a linux server, I haven't been able to test fe7481f yet to confirm everything is working, although the couple tests I have run locally on my Mac seem to work fine.

jmtsuji:
Closing this now that #63 is merged.

PR #63, opened 2023-10-20
End repair

jmtsuji:
Addresses #58 so that `repair.py` is installed as part of rotary and is integrated into the snakemake workflow. Also includes usage of TIGRFAM HMMs.

@LeeBergstrand would you mind taking a look before this PR is merged? (Thanks so much for your support for the install of repair.py!!)


LeeBergstrand:
@jmtsuji I added https://github.com/jmtsuji/rotary/pull/63/commits/e4dceffc063a47012023babe7731a423ff849564 to add the dependencies to the other requirements files. So, three files require dependency updates.

- ```environment.yaml``` - used for creating conda enviroments.
- ```meta.yaml``` - used by ```conda build``` when building conda packges.
- ```pyproject.toml``` - used by ```pip install``` when installing via ```pip``. It can only install pure Python packages offPypii (so no bioinformatics tools).

You could technically remove ```environment.yaml``` once the package is uploaded to Anaconda as ```conda build``` would specify these dependencies. You can programmatically get the dependencies out of the ```environment.yaml``` and have them templated during the build process (see https://stackoverflow.com/questions/60125363/conda-meta-yaml-requirements-run-section-dependencies-from-file)

The ```conda build``` process works like this:

1. ```Conda build``` takes the ```meta.yaml``` file and uses it to start building the Conda package.
2. ```Conda build``` installs the dependencies listed in the ```meta.yaml``` file.
3. ```Conda build``` calls ```pip install .``` to install the ```rotary``` python package.
4.  ```pip install``` uses the the dependencies in ```pyproject.toml``` and attempts to install them. However, these were already installed by ```conda build,` so it skips this step. 
5. ```pip install``` uses the the scripts list in ```pyproject.toml``` to install the CLI scripts.
6. All the resulting dependencies and files are baked by ```conda build``` into a conda bundle specific to the operating system the package was built on.
7. You then run ```conda convert``` to convert the package you built for your OS to versions that work on other operating systems (this is why, on anaconda.org, you see multiple package builds for different OSes). If the conversion fails, you must run ```conda build``` on a different OS. For example, if the Windows conversion build fails, you must run ```conda build``` on a Windows computer and upload it separately.
8. You then use the ```anaconda``` CLI tool to upload these packages to anaconda.org.

The big projects run these steps using GitHub Actions (https://docs.github.com/en/actions) to do the build and upload steps.

rotary/repair.py
```diff
@@ -748,23 +783,31 @@ def run_end_repair(long_read_filepath, assembly_fasta_filepath, assembly_info_fi
...
+    logger.info('#### Final stats: ####')
+    logger.info(f'Circular, repaired: {len(repaired_contig_names)} contig(s)')
+    logger.info(f'Circular, not repairable: {len(failed_contig_names)} contig(s)')
+    logger.info(f'Linear: {len(linear_contig_names)} contig(s)')
+    logger.info('######################')
+    logger.info(f'Output contigs are saved at {end_repaired_contigs_filepath}. '
+                f'A summary of repair work is saved at {end_repair_status_filepath}.')
 
 
 def main():
```
LeeBergstrand, line 813: @jmtsuji I moved move ```main()``` to the top of the file  justbelow the imports. That way, when the user opens the file, they don't have to scroll around searching for ```main()```. I would then order roughly where the functions occur in the file in the order called by ```main()```. Having them in that order makes it easier to look up if you read through the ```main()``` code. The exception to this pattern is that I would leave the argsparse code at the bottom.

LeeBergstrand: The order in which the functions are defined doesn't matter in Python, as they are all loaded at runtime.

rotary/repair.py
```diff
@@ -15,14 +15,15 @@
...
 # GLOBAL VARIABLES
-SCRIPT_VERSION = '0.2.0'
 DEPENDENCY_NAMES = ['flye', 'minimap2', 'samtools', 'circlator']
 
 # Set up the logger
-logging.basicConfig(format='[ %(asctime)s UTC ]: %(module)s: %(funcName)s: %(levelname)s: %(message)s')
-logging.Formatter.converter = time.gmtime
 logger = logging.getLogger(__name__)
-
+logger.setLevel(logging.DEBUG)
```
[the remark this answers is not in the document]
LeeBergstrand: @jmtsuji Okay you already do this with the ```--verbose``` flag. Great job! I would still set the default flag to info rather than debug. P.s. As discussed in other comments, I would have seen this flag earlier if ```main()``` was at the top of the file.

rotary/repair.py
```diff
@@ -780,29 +823,29 @@ def main():
...
+                     f'--overwrite flag at your own risk if you want to use an existing directory.')
+
+        raise FileExistsError
+
+    os.makedirs(args.output_dir, exist_ok=True)
+
+    # Start log file in the output dir
+    file_handler = logging.FileHandler(filename=os.path.join(args.output_dir, 'repaired.log'), mode='w')
+    file_handler.setFormatter(formatter)
+    file_handler.setLevel(logging.DEBUG)
```
LeeBergstrand, line 847: As discussed in other comments, one wants to write the info-level logs, not debug logs, by default. This way, your log files don't get massive by printing everything. So your ```--verbose``` flag should control this log level to.

rotary/repair.py
```diff
@@ -678,10 +688,30 @@ def stitch_all_contigs(circular_contig_tmp_fasta, bam_filepath, linking_outdir_b
...
+    Runs the end repair workflow
+
+    :param long_read_filepath: path to the QC-passing Nanopore reads, FastQ format; gzipped is OK
+    :param assembly_fasta_filepath: path to the assembled contigs output by Flye, FastA format
+    :param assembly_info_filepath: path to assembly_info.txt output by Flye
+    :param assembly_info_type: 'flye' type for assembly_info.txt from flye; 'custom' for custom format (see parse_cli
+                               for custom format details)
+    :param output_dir: path to the directory to save output files
+    :param length_thresholds: list of bp regions around the contig ends to attempt to subset for the assembly
+    :param keep_failed_contigs:
```
LeeBergstrand, line 701: missing parameter description.

LeeBergstrand (changes requested):

LeeBergstrand:
Great job putting this together! Looks like it's coming together splendidly. I left a lot of comments, so keep going.

Okay, right now, ```repair.py``` has this design pattern where you pass down the ```dependency_dict``` to all the child functions that call subprocesses. This method could be more straightforward to a degree.

Right now, your code is like the following:

```python
samtools_sort_args = [dependency_dict['samtools'], 'sort', '-@', str(threads), '-m', f'{threads_mem_mb}M']
subprocess.run(samtools_sort_args, check=True, input=samtools_view.stdout, stdout=bam_handle, stderr=logfile_handle)
```

I want to move your code to using ```shell=true```. This runs the sub-command in a shell and thus can call the underly programs like samtools without a path. There are security issues with ```shell=true```, but they don't apply to us because we already run this tool from CLI.

LeeBergstrand:
What are your thoughts on removing the CLI dependency check from the code? If the dependency isn't there the code would just fail and it would dump that the shell can't find it. Installing through conda should guarantee that the dependencies are installed 98% of the time.

rotary/repair.py
```diff
@@ -811,47 +854,50 @@ def main():
...
+    threads_mem_mb = int(args.threads_mem * 1024)
 
     # Startup messages
     logger.info('Running ' + os.path.basename(sys.argv[0]))
-    logger.info('Version: ' + SCRIPT_VERSION)
     logger.info('### SETTINGS ###')
-    logger.info(f'Long read filepath: {long_read_filepath}')
-    logger.info(f'Assembly FastA filepath: {assembly_fasta_filepath}')
-    logger.info(f'Assembly info filepath: {assembly_info_filepath}')
+    logger.info(f'Long read filepath: {args.long_read_filepath}')
```
LeeBergstrand, line 872: A lot of these log statements could probably be moved to the debug level.

rotary/repair.py
```diff
@@ -213,7 +211,6 @@ def map_long_reads(contig_filepath: str, long_read_filepath: str, output_bam_fil
 
     with open(log_filepath, write_mode) as logfile_handle:
         with open(output_bam_filepath, 'w') as bam_handle:
-
             # TODO - add support for different flags like -ax for pacbio
             minimap_args = [dependency_dict['minimap2'], '-t', str(threads), '-ax', 'map-ont', contig_filepath,
                             long_read_filepath]
```
LeeBergstrand, line 216: Many of your ```subprocess.run``` calls have similar code. They could be refactored into a function that takes the CLI argument string/list and the error logging file handles and then runs ```subprocess.run```  inside. For example, if this was already refactored this way we would have only one place to add ```shell=true```.

LeeBergstrand:
The following code should be moved to a function as it is repeated multiple times.

```python
 if append_log is True:
        write_mode = 'a'
    elif append_log is False:
        write_mode = 'w'
    else:
        raise ValueError
```

jmtsuji:
> @jmtsuji I added https://github.com/jmtsuji/rotary/commit/e4dceffc063a47012023babe7731a423ff849564 to add the dependencies to the other requirements files. So, three files require dependency updates.

@LeeBergstrand Thanks for the explanation and for updating the other requirement files! I'll keep this in mind when updating dependencies in the future. Your overview of how the pip and conda installers interact is very helpful.

rotary/rules/rotary.smk
```diff
@@ -8,10 +8,10 @@ import pandas as pd
 import itertools
```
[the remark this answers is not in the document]
jmtsuji: I'm also using Pycharm. Actually, I just checked, and I got the same feedback from Pycharm that those three dependencies can be removed. (Usually Pycharm auto-hides the import area, so I forget to check that part of the code... nice to know about the 'optimize imports' function.)

I think that Pycharm's support for snakefiles was not good when I started using Pycharm for snakemake projects, so Pycharm used to throw many incorrect warnings... I imagine the support has improved now. Anyway, I have cleaned up the imports section for this file and for `repair.py` in commit a491196. In future, that sounds good to use Pycharm to check for additional things to clean up in the code.

rotary/repair.py
jmtsuji: Thanks for the advice -- main() is now at the top of the file

rotary/repair.py
jmtsuji: Parameter section added

rotary/repair.py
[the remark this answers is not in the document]
jmtsuji: @LeeBergstrand Interestingly, I found that the default logger level needs to be set to the most verbose level that you want to be able to set the log handlers to. See the flow chart here: https://docs.python.org/3/howto/logging.html#logging-flow

The repair code has two log handlers: `StreamHandler()` that writes to screen and `FileHandler()` that writes to a file. As shown in the flow chart, if you set the "root" logger level to INFO, it will drop anything at DEBUG level before it has a chance to get to either of the handlers. I tested this, and indeed, you can't log at DEBUG (even if you set the handlers to DEBUG level) if the "root" logger level is INFO.

With the current setup in the repair code, the user can control the verbosity level that gets printed to the screen (via `StreamHandler()`) using the `--verbose` flag. I forced the log file (via `FileHandler()`) to always write at DEBUG in case the user wanted to check on details of the run at some point in the future. Because of this, it made sense to hard-code the "root" logger level to DEBUG outside of main().

Do you think it would be better to let the user control the verbosity of both the screen output and the file log via the `--verbose` flag? If so, then it might be possible to move the "root" logger level setting into main() so that it doesn't need to be hard-coded as DEBUG. I hope this somewhat convoluted explanation makes sense. Thanks!

rotary/repair.py
jmtsuji: See conversation about line 25 above

rotary/repair.py
jmtsuji: Done

jmtsuji:
> The following code should be moved to a function as it is repeated multiple times...

Moved to a function

jmtsuji:
> Okay, right now, repair.py has this design pattern where you pass down the dependency_dict to all the child functions that call subprocesses. This method could be more straightforward to a degree...
> I want to move your code to using shell=true. This runs the sub-command in a shell and thus can call the underlying programs like samtools without a path. There are security issues with shell=true, but they don't apply to us because we already run this tool from CLI.

@LeeBergstrand Actually, based on my previous testing, I am pretty sure that I can already call the underlying programs like samtools without a path, even if shell = False. However, I decided to explicitly set the path to the programs to avoid any weird PATH conflict issues that might happen. Do you think that such path conflicts are basically impossible? If so, I can simplify the code.

jmtsuji:
> What are your thoughts on removing the CLI dependency check and the dependancy_dict from the code? If the dependency isn't there, the code would fail, and it would dump that the shell can't find it. Installing through conda should guarantee that the dependencies are installed 98% of the time.

@LeeBergstrand I like the CLI dependency check because it tells the user upfront that a dependency is missing. Otherwise, they might have to run 50% of the pipeline before they run into the missing dependency issue. In future, I am thinking of using the dependency check function to also grab the version of the dependency tool to add to the log. So I would be happy to leave the function in the code... what are your thoughts?

rotary/repair.py
jmtsuji: @LeeBergstrand There are enough variations within the `subprocess.run` calls (e.g., where piping is used, like in `map_long_reads`, or where STDERR is pushed to STDOUT in `run_circlator_merge`, or where the program is allowed to fail in `run_flye`) that I felt it was better to not summarize these `subprocess.run` calls in a single function. I was concerned that a summary function would need to have lots of arguments (and could be confusing to understand) in order to meet the needs of each of the invocations of `subprocess.run`. But maybe it would not be as bad as I imagine? Let me know your thoughts.

jmtsuji:
> Great job putting this together! Looks like it's coming together splendidly. I left a lot of comments, so keep going.

@LeeBergstrand Thanks! Appreciate the comments. I left several responses and some unresolved threads, so please have a look and let me know your thoughts... we can move forward toward merging this PR based on your feedback.

rotary/rules/rotary.smk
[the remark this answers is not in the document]
LeeBergstrand: @jmtsuji I am using the Pycharm snakemake plugin: https://plugins.jetbrains.com/plugin/11947-snakecharm. Thats probably why I'm getting the proper error and warning messages.

rotary/repair.py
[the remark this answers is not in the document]
LeeBergstrand: @jmtsuji Thanks for the info regarding setting the initial log level to debug at the top and then reducing it later. Great to know.

**Context**: So at the production scale, you would have something like Rotary run in response to a cron job or a push notification. So, for example, when a new genome gets uploaded to the data server or Google Cloud, it would automatically run Rotary from a script on that genome, or you would rerun the entire catalogue of genomes every six months with the latest databases. This is an approach that MLBS might be taking down the road and something we are already doing with 16S seq. If a program is left in debug mode, you would be surprised how much space log files take and how long it takes to analyze them. At scale, the companies start sending logs back to a centralized server where they can be analyzed: https://www.datadoghq.com/blog/python-logging-best-practices/. You can monitor these centralized logs to get performance information from the time stamps (how long tasks take in deployment). In addition, in production, most log files are set up to roll over at a fixed amount of space or time, so they will only hold the last few days of logging. So, a lowered level of logging can lead to more days of logging stored for analysis. The standard operating procedure is leaving both the visual and written logs non-verbose, where they get logged back to files or a central server. If something goes wrong, you can go through the logs to find the failed data, and then the investigation team can download the failed data and rerun the pipeline in verbose mode to figure out what went wrong and apply fixes. The log messages should have the file names or other identifying parameters used to determine which files failed to support such investigations.

> Do you think it would be better to let the user control the verbosity of both the screen output and the file log via the `--verbose` flag? If so, then it might be possible to move the "root" logger level setting into main() so that it doesn't need to be hard-coded as DEBUG. I hope this somewhat convoluted explanation makes sense. Thanks!

Yes, I think using the ```--verbose``` flag should control the log level of both the screen output and the log file.

LeeBergstrand:
> @LeeBergstrand Actually, based on my previous testing, I am pretty sure that I can already call the underlying programs like samtools without a path, even if shell = False. However, I decided to explicitly set the path to the programs to avoid any weird PATH conflict issues that might happen. Do you think that such path conflicts are basically impossible? If so, I can simplify the code.

Given the installation, these pathing issues are generally unlikely. Because ```check_dependency()``` uses ```shutil.which(dependency_name)```, you will likely pass any pathing issues onto the Python code.  I believe the ```shutil.which``` would be affected by shell variables, symlinks, etc., including incorrect ones. In Unix, executables are files with some file system flags changed, and globally accessible executables are those added to the ```$PATH```. The default ```$PATH``` for your user should include ```/home/YOUR-USER-NAME/bin``` (i.e. ```~/bin```). Most executables are symlinked to ```~/bin```, which lets them show up in your shell when you call them.

Given the above I think we can simplify the code.

LeeBergstrand:
> > What are your thoughts on removing the CLI dependency check and the dependancy_dict from the code? If the dependency isn't there, the code would fail, and it would dump that the shell can't find it. Installing through conda should guarantee that the dependencies are installed 98% of the time.
> 
> @LeeBergstrand I like the CLI dependency check because it tells the user upfront that a dependency is missing. Otherwise, they might have to run 50% of the pipeline before they run into the missing dependency issue. In future, I am thinking of using the dependency check function to also grab the version of the dependency tool to add to the log. So I would be happy to leave the function in the code... what are your thoughts?

If the end repair pipeline takes a while, I would keep the dependency check-in. I think logging the dependency versions would be helpful.

jmtsuji:
>Given the installation, these pathing issues are generally unlikely. Because check_dependency() uses shutil.which(dependency_name), you will likely pass any pathing issues onto the Python code. I believe the shutil.which would be affected by shell variables, symlinks, etc., including incorrect ones. In Unix, executables are files with some file system flags changed, and globally accessible executables are those added to the $PATH. The default $PATH for your user should include /home/YOUR-USER-NAME/bin (i.e. ~/bin). Most executables are symlinked to ~/bin, which lets them show up in your shell when you call them.
> Given the above I think we can simplify the code.

OK, got it. I have simplified the code!

rotary/repair.py
LeeBergstrand: @jmtsuji I just created https://github.com/jmtsuji/rotary/pull/66 to do the above refactoring. Minor edits could be made. For example, moving the ```logger debug``` into a function with the ```subprocess.run``` call so the log is not called multiple times. This way, you can change the debug logger statement without changing all eight of them manually.

jmtsuji:
End repair might take several minutes (or longer if a large assembly is input), so let's leave `dependency_check()` in the code.

rotary/rules/rotary.smk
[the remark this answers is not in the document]
jmtsuji: Thanks for the tip -- I will check this out!

rotary/repair.py
jmtsuji: Thanks -- I reviewed and have approved the current version. I have to run to a (likely) all-day meeting now, so I cannot look into details further, but feel free to continue to work on this today if you would like.

jmtsuji: I think I've covered all other needed code changes, so I am also OK with you merging (or approving merge) of this PR if you think it's ready enough for now.

rotary/repair.py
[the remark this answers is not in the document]
jmtsuji: This is really helpful additional context -- thanks! I have updated the code as discussed.

rotary/repair.py
jmtsuji: (Or feel free to leave further TODOs on my end)

jmtsuji: I'll resolve this conversation now that #66 is merged.

LeeBergstrand:
@jmtsuji We are pretty close to being able to merge this in—one last issue other than removing the ```dependency_dict```,

https://github.com/jmtsuji/rotary/blob/0b984a5dd01d80f999d0f688c9aac3fb4c228ffe/rotary/repair.py#L463

Is this the correct function argument mapping? In the other runs commands, you mapped ```stderr``` to the log file. These arguments are mapped correctly?

jmtsuji:
>Is this the correct function argument mapping? In the other runs commands, you mapped stderr to the log file. These arguments are mapped correctly?

@LeeBergstrand Yes, this is correct. This particular tool writes the log info to STDOUT, so I had to direct STDOUT to the logfile. I wanted to retain STDERR info as well in case anything got printed there. (By the way, this is the tool that I plan to replace with in-house code).

LeeBergstrand:
> > Is this the correct function argument mapping? In the other runs commands, you mapped stderr to the log file. These arguments are mapped correctly?
> 
> @LeeBergstrand Yes, this is correct. This particular tool writes the log info to STDOUT, so I had to direct STDOUT to the logfile. I wanted to retain STDERR info as well in case anything got printed there. (By the way, this is the tool that I plan to replace with in-house code).

Great I wanted to confirm.

jmtsuji:
@LeeBergstrand Dependency dict and logger changes are addressed. End-to-end tests are passing. OK to merge?

LeeBergstrand:
@jmtsuji Great. Feel free to merge.

PR #66, opened 2023-10-24
Refactor subcommand calls in rotary repair.py

LeeBergstrand:
Addresses: https://github.com/jmtsuji/rotary/pull/63#discussion_r1368078767

- Move ```subcommand.run``` to a function
- Move ```logger.debug()``` call before ```subcommand.run``` into the function so it can be edited in one place.
- Make ```check=True``` to be the fault behaviour when calling a subcommand by moving it into the function. Now it be left out.
- Move ```std.stdout``` into the function so one can pass the output from the prior command directly into the subcommand call.

### DIFF
--- a/rotary/repair.py
+++ b/rotary/repair.py
@@ -319,24 +319,39 @@ def map_long_reads(contig_filepath: str, long_read_filepath: str, output_bam_fil
             # TODO - add support for different flags like -ax for pacbio
             minimap_args = [dependency_dict['minimap2'], '-t', str(threads), '-ax', 'map-ont', contig_filepath,
                             long_read_filepath]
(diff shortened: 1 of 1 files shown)


LeeBergstrand:
@jmtsuji See: https://github.com/jmtsuji/rotary/pull/63#discussion_r1369418086

LeeBergstrand:
@jmtsuji I could only use Pycharm's automated refactoring tools (https://www.jetbrains.com/help/pycharm/refactoring-source-code.html) to do these edits. If you could go through this code in review to make sure I mapped all the commands properly that would be great. Also, once the conflicts are resolved, I would re-run your test to ensure nothing breaks.

jmtsuji (approved):
This looks great and passes end-to-end tests -- feel free to merge!

PR #69, opened 2023-10-26
Utilize sample tsv for single sample

LeeBergstrand:
- Get single sample information from the sample.tsv file rather than the config.
- Slight bug fix to rotary.py


rotary/config.yaml
```diff
@@ -1,16 +1,7 @@
...
-sample_id: "E_coli"
-# Nanopore data
-longreads: "E_coli.fastq.gz"
-# QC'ed short reads
-# Set qc_short_r1 (and qc_short_r1) to "None" (with quotes) to skip short read polishing
-qc_short_r1: "QC_R1.fastq.gz"
-qc_short_r2: "QC_R2.fastq.gz"
-#####################
+# path to the sample tsv file
+sample_tsv: "example/path"
```
jmtsuji, line 4: @LeeBergstrand I wonder if it might be possible to completely separate `samples_tsv` from the config file in the long term? This might allow the config file to be more flexibly applied to different runs. I think that ATLAS somehow avoids specifying `samples_tsv` in the config, but I am not sure how. Anyway, I think the current solution is totally OK for this PR.

jmtsuji: Also, perhaps it would be better to specify the sample path as something like "/example/samples.tsv"? This change would:
- show that an absolute path is better than a relative path (unless you have found a consistent way to work with relative paths in the config now) -- I think the relative paths I had in the config file before were confusing
- show that the samples file name is usually `samples.tsv`

What do you think?

rotary/rules/rotary.smk
```diff
@@ -5,14 +5,32 @@ import os
...
+
 VERSION_POLYPOLISH="0.5.0"
 VERSION_DFAST="1.2.18"
 VERSION_EGGNOG="5.0.0" # See http://eggnog5.embl.de/#/app/downloads
 START_HMM_NAME = os.path.splitext(os.path.basename(config.get("hmm_url")))[0]
 VERSION_GTDB_COMPLETE= "214.1" # See https://data.gtdb.ecogenomic.org/releases/
 VERSION_GTDB_MAIN=VERSION_GTDB_COMPLETE.split('.')[0] # Remove subversion
+DB_DIR_PATH = config.get('db_dir')
+
+samples = []
```
jmtsuji, line 20: I am still not that familiar with best practices for complex snakefiles. I wonder, though, it it might improve readability to put this sample parser into a function and then call the function? This isn't necessary with a single code block at this point, but it is something we should consider down the road.

rotary/rules/rotary.smk
```diff
@@ -455,8 +473,8 @@ rule polish_polypolish:
     benchmark:
         "benchmarks/{step}/polypolish.txt"
     params:
-        qc_short_r1=config.get("qc_short_r1"),
-        qc_short_r2=config.get("qc_short_r2")
+        qc_short_r1=SHORT_R1,
```
jmtsuji, line 476: One nice thing about adding the short read QC workflow in the future is that it will reduce the number of calls to the SHORT_R1 and SHORT_R2 globals substantially!

jmtsuji (commented):
Looks good to me -- congrats on integrating samples.tsv into the workflow for a single sample! I've added a couple minor comments and questions during my code review, but I don't think these necessarily require changes to the code, so I've left this review as a "Comment" rather than a "Request changes". Feel free to go ahead and merge the PR after taking a look at the comments/questions I left.

rotary/rules/rotary.smk
LeeBergstrand: @jmtsuji Done.

rotary/config.yaml
LeeBergstrand: @jmtsuji The new code looks for the sample.tsv in the current working directory (where the config is written). Thoughts?

LeeBergstrand:
@jmtsuji I made new code that addresses some of your comments and https://github.com/jmtsuji/rotary/issues/53

- added flag to creating sample objects that skips the integrity checks.
- skips integrity checks in run_one and when parsing existing sample tsvs.
- adds functions that parses the sample.tsv into dictionary of sample objects keyed by sample identifer.
- rotary.smk now uses these objects.

rotary/rotary.py
```diff
@@ -135,18 +134,16 @@ def run_one(args, config, output_dir_path, conda_env_directory, jobs, snakemake_
...
-    sequencing_files = [SequencingFile(path) for path in [long_path, left_path, right_path]]
+    sequencing_files = [SequencingFile(path) for path in [(get_cli_arg_path(args, 'long')),
+                                                          (get_cli_arg_path(args, 'left')),
+                                                          (get_cli_arg_path(args, 'right'))]]
+
     sample = Sample(long_file=sequencing_files[0],
                     short_file_one=sequencing_files[1],
-                    short_file_two=sequencing_files[2])
+                    short_file_two=sequencing_files[2],
+                    integrity_check=False) # Don't do integrity check on user specified files.
```
jmtsuji, line 144: I like this update to simplify the integrity check for run_one. Based on a quick look at the code, my guess is that run_one will start the run even if not enough input files are supplied (e.g., if there is no long read file, or if only R1 or only R2 are supplied). This could potentially lead to cryptic error messages for the user. In the long term, we could consider adding some code to make error reporting more clear when sufficient input files are supplied (if the code doesn't already catch this), but I don't think this is needed at the moment.

rotary/config.yaml
jmtsuji: This sounds OK to me -- thanks!

jmtsuji: We should add this info to the CLI printout for `rotary run` for clarity at some point I suppose.

jmtsuji (approved):
Thanks for these updates! Nice that the sample TSV parser can be moved to samples.py. The updated code looks good to me, although see my two brief comments that are more on the UI end. These comments could be moved to issues or commited to the PR before the PR is merged. Feel free to resolve those two conversations and then merge this PR.

rotary/rotary.py
LeeBergstrand: I believe that there is already code for this but in a different area.

LeeBergstrand: @jmtsuji Should be more clearly addressed in https://github.com/jmtsuji/rotary/pull/70.

PR #70, opened 2023-10-27
Refactor rotary cli

LeeBergstrand:
- refactor ```rotary.py``` for improved reliability and expandability
- move all code that processes sub-commands below the switch case that selects a sub-command. This should support the easy addition of sub-commands from other modules. This should support the addition of a new command to support https://github.com/jmtsuji/rotary/issues/60
- concentrated code for parsing CLI arguments in sub-command functions or shared between sub-command functions. This made it much easier to fix various bugs regarding where and when to generate output files.

### DIFF
--- a/rotary/rotary.py
+++ b/rotary/rotary.py
@@ -14,7 +14,7 @@ import psutil
 from ruamel.yaml import YAML
 
 from rotary.sample import SequencingFile, Sample, is_fastq_file, create_sample_tsv
-from rotary.utils import get_cli_arg_path
+from rotary.utils import get_cli_arg_path, get_cli_arg, check_for_files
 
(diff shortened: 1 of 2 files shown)


LeeBergstrand:
@jmtsuji I went to do some bug fixes in rotary.py and decided it would be worthwhile to refactor it to make fixing the bugs more straightforward and to make it easier to ensure that the correct behaviour is occurring.

jmtsuji (approved):

jmtsuji:
@LeeBergstrand Nice refactoring work. I like the simplified way that the modules are handled now -- should make it easier to integrate other modules in the future, I agree. Looks good to me, so feel free to merge this PR!

PR #71, opened 2023-11-03
Add multi sample capability.

LeeBergstrand:
Adds preliminary multi-sample capability.

### DIFF
--- a/rotary/rules/rotary.smk
+++ b/rotary/rules/rotary.smk
@@ -117,7 +117,7 @@ rule download_eggnog_db:
     shell:
         """
         mkdir -p {params.db_dir}
-        download_eggnog_data.py -y --data_dir {params.db_dir} > {log} 2>&1
+        download_eggnog_data.py -y -M --data_dir {params.db_dir} > {log} 2>&1
         touch {output.install_finished}
         """
 
@@ -224,22 +224,22 @@ rule set_up_sample_directories:
     input:
         SAMPLE_TSV_PATH
     output:
-        long_reads = expand("{sample}/{sample}_long.fastq.gz", sample=SAMPLE_NAMES),
-        short_R1_reads = expand("{sample}/{sample}_R1.fastq.gz", sample=SAMPLE_NAMES),
-        short_R2_reads = expand("{sample}/{sample}_R2.fastq.gz", sample=SAMPLE_NAMES),
+        long_reads = expand("{sample}/raw/{sample}_long.fastq.gz", sample=SAMPLE_NAMES),
+        short_R1_reads = expand("{sample}/raw/{sample}_R1.fastq.gz", sample=SAMPLE_NAMES),
+        short_R2_reads = expand("{sample}/raw/{sample}_R2.fastq.gz", sample=SAMPLE_NAMES),
     run:
         for sample in SAMPLES.values():
             identifier = sample.identifier
-            symlink_or_compress(sample.long_read_path,f'{identifier}/{identifier}_long.fastq.gz')
-            symlink_or_compress(sample.short_read_left_path,f'{identifier}/{identifier}_R1.fastq.gz')
-            symlink_or_compress(sample.short_read_right_path,f'{identifier}/{identifier}_R2.fastq.gz')
+            symlink_or_compress(sample.long_read_path,f'{identifier}/raw/{identifier}_long.fastq.gz')
+            symlink_or_compress(sample.short_read_left_path,f'{identifier}/raw/{identifier}_R1.fastq.gz')
+            symlink_or_compress(sample.short_read_right_path,f'{identifier}/raw/{identifier}_R2.fastq.gz')
 
 
 rule nanopore_qc_filter:
     input:
-        "{sample}/{sample}_long.fastq.gz"
+        "{sample}/raw/{sample}_long.fastq.gz"
     output:
-        "{sample}/qc_long/nanopore_qc.fastq.gz"
+        "{sample}/qc_long/{sample}_nanopore_qc.fastq.gz"
     conda:
         "../envs/mapping.yaml"
     log:
@@ -262,9 +262,9 @@ rule nanopore_qc_filter:
 
 rule qc_long_length_hist:
     input:
-        "{sample}/qc_long/nanopore_qc.fastq.gz"
+        "{sample}/qc_long/{sample}_nanopore_qc.fastq.gz"
     output:
-        "{sample}/qc_long/length_hist.tsv"
+        "{sample}/qc_long/{sample}_length_hist.tsv"
     conda:
         "../envs/mapping.yaml"
     log:
@@ -284,9 +284,9 @@ rule qc_long_length_hist:
 
 rule qc_long_length_stats:
     input:
-        "{sample}/qc_long/length_hist.tsv"
+        "{sample}/qc_long/{sample}_length_hist.tsv"
     output:
-        "{sample}/stats/qc_long_length_stats.txt"
+        "{sample}/stats/{sample}_qc_long_length_stats.txt"
     log:
         "{sample}/logs/qc/qc_long_length_stats.log"
     benchmark:
@@ -314,17 +314,17 @@ rule qc_long_length_stats:
 
 rule qc_long:
     input:
-        expand("{sample}/stats/qc_long_length_stats.txt",sample=SAMPLE_NAMES)
+        expand("{sample}/stats/{sample}_qc_long_length_stats.txt",sample=SAMPLE_NAMES)
     output:
         temp(touch("checkpoints/qc_long"))
 
 
 rule assembly_flye:
     input:
-        "{sample}/qc_long/nanopore_qc.fastq.gz"
+        "{sample}/qc_long/{sample}_nanopore_qc.fastq.gz"
     output:
-        assembly="{sample}/assembly/flye/assembly.fasta",
-        info="{sample}/assembly/flye/assembly_info.txt",
+        assembly="{sample}/assembly/flye/{sample}_assembly.fasta",
+        info="{sample}/assembly/flye/{sample}_assembly_info.txt",
         output_dir=directory("{sample}/assembly/flye")
     conda:
         "../envs/assembly_flye.yaml"
@@ -342,7 +342,9 @@ rule assembly_flye:
     shell:
         """
         flye --{params.input_mode} {input} {params.read_error} --out-dir {output.output_dir} {params.meta_mode} \
-          --iterations {params.polishing_rounds} -t {threads} > {log} 2>&1
+          --iterations {params.polishing_rounds} -t {threads} > {log} 2>&1 
+        mv {output.output_dir}/assembly.fasta {output.assembly} 
+        mv {output.output_dir}/assembly_info.txt {output.info} 
         """
 
 
@@ -350,12 +352,12 @@ rule assembly_flye:
 # TODO - the math for memory per thread should really be done somewhere other than 'resources' - what is best practice?
 rule assembly_end_repair:
     input:
-        qc_long_reads="{sample}/qc_long/nanopore_qc.fastq.gz",
-        assembly="{sample}/assembly/flye/assembly.fasta",
-        info="{sample}/assembly/flye/assembly_info.txt"
+        qc_long_reads="{sample}/qc_long/{sample}_nanopore_qc.fastq.gz",
+        assembly="{sample}/assembly/flye/{sample}_assembly.fasta",
+        info="{sample}/assembly/flye/{sample}_assembly_info.txt"
     output:
-        assembly="{sample}/assembly/end_repair/repaired.fasta",
-        info="{sample}/assembly/end_repair/repaired_info.tsv",
+        assembly="{sample}/assembly/end_repair/{sample}_repaired.fasta",
+        info="{sample}/assembly/end_repair/{sample}_repaired_info.tsv",
         output_dir=directory("{sample}/assembly/end_repair"),
     log:
         "{sample}/logs/assembly/end_repair.log"
@@ -391,16 +393,18 @@ rule assembly_end_repair:
           --overwrite \
           {params.keep_going} \
           > {log} 2>&1
+        mv {output.output_dir}/repaired.fasta {output.assembly} 
+        mv {output.output_dir}/repaired_info.tsv {output.info}
         """
 
 
 rule finalize_assembly:
     input:
-        assembly="{sample}/assembly/end_repair/repaired.fasta",
-        info="{sample}/assembly/end_repair/repaired_info.tsv"
+        assembly="{sample}/assembly/end_repair/{sample}_repaired.fasta",
+        info="{sample}/assembly/end_repair/{sample}_repaired_info.tsv"
     output:
-        assembly="{sample}/assembly/assembly.fasta",
-        info="{sample}/assembly/circular_info.tsv"
+        assembly="{sample}/assembly/{sample}_assembly.fasta",
+        info="{sample}/assembly/{sample}_circular_info.tsv"
     run:
         source_relpath = os.path.relpath(str(input.assembly),os.path.dirname(str(output.assembly)))
         os.symlink(source_relpath,str(output.assembly))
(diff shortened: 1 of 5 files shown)


LeeBergstrand:
@jmtsuji Hey, could you review this visually / on your server and see if you can find any bugs. I assume there will be when I finally run it, I'm updating my test server environment tomorrow but I wanted you to give a heads up for review to pick out any big errors.

```--dry-run``` works and makes the right copies of rules to run.

**Comment by LeeBergstrand** [comment]
So at the top of the snakemake file I create a dictionary of sample objects, keyed by identifiers, representing each sample.

I also assigned a list at the top called SAMPLE_NAMES that contains the sample identifier list.

I added a rule called ```set_up_sample_directories``` that takes this dictionary and uses it to create all the sample output directories and symlink the input fastq files to directories (if they aren't gzipped I gzip them here).

for this rule and each of the checkpoint rules I used expand to get the list of output files based on the sample identifiers.

```python
expand("{sample}/{sample}_long.fastq.gz", sample=SAMPLE_NAMES) # This expands out to a list of long fastq files paths for each sample.
```

I added `{sample}/` wild cards to all the paths.

for some of your existing ```expand()``` I added ```{{sample}}/```. the second braces tells the expand not to fill in ```sample```. So you get an expansion but each expanded path starts with `{sample}/`which is later interpreted as a wildcard.

In rules ```params:``` cannot access wildcards directly so I either moved them into input or output or manually accessed the wild card with a lambda function.

```python
strain=lambda wildcards: wildcards.sample # Get sample name from wildcards.
```

I not sure if this is the proper way to access them but we will see at runtime.

LeeBergstrand:
I also moved some of your checks whether or not to use short read polishing to its own variable, `POLISH_WITH_SHORT_READS`, at the top of the the file. I think we could add this as a config variable down the road.

I estimate that it might be trick to figure out the logic of short read polish or not on a per sample basis would be tricky and might need a significant rewrite of the switching logic.

rotary/rules/rotary.smk
```diff
@@ -501,40 +524,38 @@ rule polish_polypolish:
...
+        qc_short_r1 = "{sample}/{sample}_R1.fastq.gz",
+        qc_short_r2 = "{sample}/{sample}_R2.fastq.gz",
+        polished = "{sample}/polish/polypolish/polypolish.fasta"
     output:
-        "polish/polca/polca.fasta",
-        temp("polish/polca/polypolish.fasta.unSorted.sam"),
-        temp("polish/polca/polypolish.fasta.alignSorted.bam")
+        polca_output = "{sample}/polish/polca/polca.fasta",
+        polypolish_sam = temp("{sample}/polish/polca/polypolish.fasta.unSorted.sam"),
+        polypolish_bam = temp("{sample}/polish/polca/polypolish.fasta.alignSorted.bam"),
```
jmtsuji, line 533: Minor typo - I think an extra comma is here that should be removed

rotary/rules/rotary.smk
```diff
@@ -501,40 +524,38 @@ rule polish_polypolish:
...
         config.get("threads",1)
     resources:
         mem=config.get("memory")
     shell:
         """
-        cd {params.outdir}
-        polca.sh -a ../../{input} -r "{params.qc_short_r1} {params.qc_short_r2}" -t {threads} -m {resources.mem}G > ../../{log} 2>&1
-        ln -s "polypolish.fasta.PolcaCorrected.fa" "polca.fasta"
-        cd ../..
+        polca.sh -a {input.polished} -r "{input.qc_short_r1} {input.qc_short_r2}" -t {threads} -m {resources.mem}G > {log} 2>&1
```
jmtsuji, line 546: Sadly, POLCA dumps its output files in the directory where it is run. As of the last version I checked, there is no way to specify the output directory. This is why my old rule had some odd code in it to `cd` to the output directory before running the command. Unless the POLCA UI has been updated, we might have to keep that odd logic in the code

rotary/rules/rotary.smk
```diff
@@ -1003,29 +1022,29 @@ rule circularize:
...
-        "logs/annotation/annotation_dfast.log"
+        "{sample}/logs/annotation/annotation_dfast.log"
     benchmark:
-        "benchmarks/annotation/annotation_dfast.txt"
+        "{sample}/benchmarks/annotation/annotation_dfast.txt"
     params:
-        outdir="annotation/dfast",
         db=directory(os.path.join(DB_DIR_PATH,"dfast_" + VERSION_DFAST)),
-        strain=SAMPLE_ID
+        strain=lambda wildcards: wildcards.sample # Get sample name from wildcards.
```
jmtsuji, line 1039: I'm not sure if this will work, but is it possible to simplify this code and just use `{sample}`? See https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#non-file-parameters-for-rules

rotary/rules/rotary.smk
```diff
@@ -1039,91 +1058,89 @@ rule run_dfast:
...
-        "logs/annotation/gtdbtk.log"
+        "{sample}/logs/annotation/gtdbtk.log"
     benchmark:
-        "benchmarks/annotation/gtdbtk.txt"
+        "{sample}/benchmarks/annotation/gtdbtk.txt"
     params:
-        outdir="annotation/gtdbtk/run_files",
         db=directory(os.path.join(DB_DIR_PATH,"GTDB_" + VERSION_GTDB_COMPLETE)),
-        genome_id=SAMPLE_ID,
+        genome_id=lambda wildcards: wildcards.sample, # Get sample name from wildcards.
```
jmtsuji, line 1104: As above, I wonder if {sample} alone is OK

rotary/rules/rotary.smk
```diff
@@ -1163,13 +1180,13 @@ rule calculate_final_long_read_coverage:
...
+        long_read_coverage="{sample}/annotation/coverage/long_read_coverage.tsv",
     output:
-        logs=temp(directory("annotation/logs")),
-        stats=temp(directory("annotation/stats"))
+        logs=temp(directory("{sample}/annotation/logs")),
+        stats=temp(directory("{sample}/annotation/stats"))
     params:
-        logs="logs",
-        stats="stats"
+        logs=lambda wildcards: f"{wildcards.sample}/logs",
```
jmtsuji, line 1188: Possible to simplify via {sample}? (See above)

rotary/rules/rotary.smk
```diff
@@ -1183,19 +1200,19 @@ rule symlink_logs:
...
+        "{sample}/annotation/stats"
     output:
-        "summary.zip"
+        "{sample}/summary.zip"
     log:
-        "logs/annotation/summarize_annotation.log"
+        "{sample}/logs/annotation/summarize_annotation.log"
     params:
-        zipdir="annotation"
+        zipdir=lambda wildcards: f"{wildcards.sample}/annotation"
```
jmtsuji, line 1215: Possible to simplify to {sample}? (see above)

rotary/rules/rotary.smk
```diff
@@ -221,17 +217,45 @@ rule build_gtdb_mash_ref_database:
...
 
+rule set_up_sample_directories:
+    input:
+        "samples.tsv"
+    output:
+        long_reads = expand("{sample}/{sample}_long.fastq.gz", sample=SAMPLE_NAMES),
+        short_R1_reads = expand("{sample}/{sample}_R1.fastq.gz", sample=SAMPLE_NAMES),
+        short_R2_reads = expand("{sample}/{sample}_R2.fastq.gz", sample=SAMPLE_NAMES),
+    run:
+        def symlink_or_compress(in_file_path, out_file_path):
```
jmtsuji, line 228: I wonder if this function could be moved to utils -- what do you think?

rotary/rules/rotary.smk
```diff
@@ -300,25 +324,25 @@ rule qc_long_length_stats:
...
 rule assembly_flye:
     input:
-        "qc_long/nanopore_qc.fastq.gz"
+        "{sample}/qc_long/nanopore_qc.fastq.gz"
     output:
-        "assembly/flye/assembly.fasta",
-        "assembly/flye/assembly_info.txt"
+        assembly="{sample}/assembly/flye/assembly.fasta",
+        info="{sample}/assembly/flye/assembly_info.txt",
+        output_dir=directory("{sample}/assembly/flye")
```
jmtsuji, line 338: Now that the path contains a wildcard, I support this move of output directory paths to the `output` field -- this is better practice I think. Similarly, thanks for moving the short read input files (in other rules) from the params to the input field

rotary/rules/rotary.smk
```diff
@@ -221,17 +217,45 @@ rule build_gtdb_mash_ref_database:
         """
 
 
+rule set_up_sample_directories:
+    input:
+        "samples.tsv"
+    output:
+        long_reads = expand("{sample}/{sample}_long.fastq.gz", sample=SAMPLE_NAMES),
```
jmtsuji, line 224: I think it might help for file organization to move these raw input files to a subfolder within {samples}, e.g., `{samples}/raw`. We might also consider setting the `temp()` flag on these. What do you think?

rotary/rules/rotary.smk
```diff
@@ -20,20 +21,15 @@ DB_DIR_PATH = config.get('db_dir')
...
-SHORT_R2 = first_sample.short_read_right_path
+SAMPLE_NAMES = list(SAMPLES.keys())
 
 # Specify the minimum snakemake version allowable
 min_version("7.0")
 # Specify shell parameters
 shell.executable("/bin/bash")
 shell.prefix("set -o pipefail; ")
 
+POLISH_WITH_SHORT_READS = True
```
jmtsuji, line 32: Thanks for adding this temporary workaround. I agree, we'll need to rework the logic for short read polishing to make it work with the multi-sample workflow

rotary/rules/rotary.smk
```diff
@@ -221,17 +217,45 @@ rule build_gtdb_mash_ref_database:
         """
 
 
+rule set_up_sample_directories:
+    input:
+        "samples.tsv"
```
jmtsuji, line 222: I wonder if this would recognize the variable `sample_tsv_path` from above? If so, that might be an elegant solution to use that variable as the input rather than writing "samples.tsv" twice in the snakefile. `sample_tsv_path` might need to be changed to uppercase for clarity (since it would behave as a global variable) in that case

jmtsuji (changes requested):
@LeeBergstrand Nice progress! Great that you found a relatively simple solution to getting the sample files into the pipeline, via the SAMPLES global and some expand() functions. I wasn't able to test the code yet, but I left several comments of things you might consider incorporating into this PR. I'll try to run an end-to-end test early next week (probably Tues.), if that works, and I'll let you know how it goes.

rotary/rules/rotary.smk
jmtsuji: @LeeBergstrand For now, I think the simplest solution would be to lock a single rotary run to one run mode (with short read polishing or without). The run mode can be set based on if R1 and R2 files are in the samples.tsv file or not. Do you think this might be a reasonable target for the MVP version of rotary? (It would only take minor modifications of the current code.)

Longer-term, it would be great if we could make rotary auto-determine whether or not to use short reads on a per-sample basis. Basically, the DAG will need to go in different directions depending on the sample, and this should ideally be defined at the start of execution of the pipeline. I am not sure how to do this. The closest thing I know how to do is to set up snakemake checkpoints that dynamically adjust the DAG partway through the pipeline, based on outputs of some commands. I think these snakemake checkpoints should generally be avoided when possible, because (I think) they cause the parts of the DAG downstream from the checkpoint to be hidden (e.g., when `--dry-run` is used). It might take some time to figure out an ideal solution...

jmtsuji:
As one other note, I haven't been able to commit new rules yet for short read QC, although I have some draft code started. Sorry for the delay here! Might it work well for me to commit the short read QC rules directly to the multi-sample version of the workflow, given that the multi-sample version is already in reasonable shape? For example, next week I could make a fork of this branch and use it to add the short read QC rules. Thanks!

rotary/rules/rotary.smk
LeeBergstrand: @jmtsuji In the shell code below, I take a different approach. I run the POLCA, and it dumps the output to the CWD, as you mentioned. I then move (`mv`) these output files to the corrected directory afterwards. This avoids having to change the directory. It also allows the pathing to be easier to understand.

rotary/rules/rotary.smk
LeeBergstrand: @jmtsuji Thanks. I misunderstood this. I thought you couldn't use written wildcards in params because of some of the other examples I was seeing.

rotary/rules/rotary.smk
LeeBergstrand: @jmtsuji Does Polypolish make more output files than the ones listed?

rotary/rules/rotary.smk
LeeBergstrand: @jmtsuji I'll have to check if the sample object still enforces having both short and long reads during its instantiation.

LeeBergstrand: @jmtsuji My main concern is that some samples have short reads and others don't. You would have to solve that with by-sample switching logic, which is a rewrite. Is that situation common? I am leaning towards putting a flag in the config for now as the MVP and getting things merged into development so you can start finalizing the Illumina QC.

rotary/rules/rotary.smk
LeeBergstrand: Done.

rotary/rules/rotary.smk
LeeBergstrand: @jmtsuji Not that, in most cases, these files will be symlinked, so they won't take up any disk space. However, yes, I think moving them to a `raw` folder makes sense. I like having a pointer to the raw files for QC/Debugging purposes.

LeeBergstrand:
> As one other note, I haven't been able to commit new rules yet for short read QC, although I have some draft code started. Sorry for the delay here! Might it work well for me to commit the short read QC rules directly to the multi-sample version of the workflow, given that the multi-sample version is already in reasonable shape? For example, next week I could make a fork of this branch and use it to add the short read QC rules. Thanks!

@jmtsuji This sounds good to me. Feel free to branch and merge into this branch.

LeeBergstrand:
TODO: I also think we should add more `{sample}` wild cards to the output file names so you can differentiate files from each other. For example, name the different `assembly.fasta` files in different sample folders with their sample names (e.g. `bs4_assembly.fasta`).

rotary/rules/rotary.smk
LeeBergstrand: @jmtsuji I reverted back to your way of doing things. But I had to add `../` to all the paths to compensate for the additional `{sample}/` in the output directory move.

rotary/rules/rotary.smk
LeeBergstrand: Addressed

LeeBergstrand:
@jmtsuji New updates to this branch to fix some preliminary bug fixes I encountered.

rotary/rules/rotary.smk
LeeBergstrand: @jmtsuji This will effect your pathing for some of the rules where you `cd` back and forth.

rotary/rules/rotary.smk
jmtsuji: @LeeBergstrand OK, sounds good!

rotary/rules/rotary.smk
jmtsuji: @LeeBergstrand OK, let's use a `raw` folder then. I'll see during end-to-end testing if anything goes awry with rules that need to `cd` back and forth

rotary/rules/rotary.smk
jmtsuji: @LeeBergstrand I share your concern -- I imagine it will be fairly common that some genomes will have short reads and some won't. For now, I think the easiest MVP solution is to switch on a per-run basis between short read vs. no short read mode, e.g., via a CLI flag, like you mentioned. But longer-term, I definitely think we should try to code some by-sample switching logic.

jmtsuji:
> TODO: I also think we should add more {sample} wild cards to the output file names so you can differentiate files from each other. For example, name the different assembly.fasta files in different sample folders with their sample names (e.g. bs4_assembly.fasta).

Agreed -- this sounds like a nice idea. For now, I will focus on end-to-end testing, but let's keep this in mind when doing edits to the snakefile down the road.

rotary/rules/rotary.smk
jmtsuji: Update: I saw that you made the multi pathing branch -- thanks!!

rotary/rules/rotary.smk
```diff
@@ -17,23 +18,21 @@ START_HMM_NAME = os.path.splitext(os.path.basename(config.get("hmm_url")))[0]
...
+SAMPLE_NAMES = list(SAMPLES.keys())
 
 # Specify the minimum snakemake version allowable
 min_version("7.0")
 # Specify shell parameters
 shell.executable("/bin/bash")
 shell.prefix("set -o pipefail; ")
 
+if str(config.get('polish_with_short_reads')).lower() == 'true':
+    POLISH_WITH_SHORT_READS = True
```
jmtsuji, line 33: Nice addition for the MVP version

rotary/rules/rotary.smk
jmtsuji: Looks OK so far based on a quick code review of the multi pathing branch.

jmtsuji (commented):
Latest code looks good to me, pending an end-to-end test and merging of the multi pathing branch. Will start an end-to-end test on this branch now and let you know the results.

rotary/rules/rotary.smk
LeeBergstrand: @jmtsuji I will leave it as a config variable for now for the near-term MVP. We can add it as a command-line flag or other methods once we have everything merged into develop.

PR #73, opened 2023-11-07
Improve multi sample pathing

LeeBergstrand:
- Move input files to raw directory of each sample.
- Add sample prefixes to output files using snakemake wildcards.
- tell eggnog mapper to use sample name prefixes for its output files.

### DIFF
--- a/rotary/rules/rotary.smk
+++ b/rotary/rules/rotary.smk
@@ -117,7 +117,7 @@ rule download_eggnog_db:
     shell:
         """
         mkdir -p {params.db_dir}
-        download_eggnog_data.py -y --data_dir {params.db_dir} > {log} 2>&1
+        download_eggnog_data.py -y -M --data_dir {params.db_dir} > {log} 2>&1
         touch {output.install_finished}
         """
 
@@ -1068,7 +1068,7 @@ rule run_eggnog:
         protein="{sample}/annotation/dfast/{sample}_protein.faa",
         install_finished=os.path.join(DB_DIR_PATH,"checkpoints","eggnog_" + VERSION_EGGNOG)
     output:
-        eggnog_annotations="{sample}/annotation/eggnog/eggnog.emapper.annotations",
+        eggnog_annotations="{sample}/annotation/eggnog/{sample}.emapper.annotations",
         outdir= directory("{sample}/annotation/eggnog")
     conda:
         "../envs/eggnog.yaml"
@@ -1079,14 +1079,16 @@ rule run_eggnog:
     params:
         prefix='{sample}',
         db=directory(os.path.join(DB_DIR_PATH,"eggnog_" + VERSION_EGGNOG)),
-        sensmode=config.get("eggnog_sensmode")
+        sensmode=config.get("eggnog_sensmode"),
+        search_tool=config.get('eggnog_search_tool')
     threads:
         config.get("threads",1)
     shell:
         """
         mkdir -p {output.outdir}/tmp
-        emapper.py --cpu {threads} -i {input.protein} --itype proteins -m diamond --sensmode {params.sensmode} \
-          --dbmem --output eggnog --output_dir {output.outdir} --temp_dir {output.outdir}/tmp --output {params.prefix}  \
+        emapper.py --cpu {threads} -i {input.protein} --itype proteins -m {params.search_tool} \
+          --sensmode {params.sensmode} --dbmem --output eggnog --output_dir {output.outdir} \
+          --temp_dir {output.outdir}/tmp --output {params.prefix} \
           --data_dir {params.db} --override > {log} 2>&1
         rm -r {output.outdir}/tmp
         """
@@ -1100,7 +1102,7 @@ rule run_gtdbtk:
     output:
         batchfile=temp("{sample}/annotation/gtdbtk/batchfile.tsv"),
         annotation="{sample}/annotation/gtdbtk/{sample}_gtdbtk.summary.tsv",
-        outdir=directory("{sample}annotation/gtdbtk/run_files")
+        outdir=directory("{sample}/annotation/gtdbtk/run_files")
     conda:
         "../envs/gtdbtk.yaml"
     log:
@@ -1209,7 +1211,7 @@ rule symlink_logs:
 rule summarize_annotation:
     input:
         "{sample}/annotation/dfast/{sample}_genome.fna",
-        "{sample}/annotation/eggnog/eggnog.emapper.annotations",
+        "{sample}/annotation/eggnog/{sample}.emapper.annotations",
         "{sample}/annotation/gtdbtk/{sample}_gtdbtk.summary.tsv",
         expand("{{sample}}/annotation/coverage/{{sample}}_{type}_coverage.tsv",
             type=["short_read", "long_read"] if POLISH_WITH_SHORT_READS == True else ["long_read"]),
--- a/rotary/rules/rotary.smk
+++ b/rotary/rules/rotary.smk
@@ -1067,6 +1067,7 @@ rule run_eggnog:
     benchmark:
         "{sample}/benchmarks/annotation/eggnog.txt"
     params:
+        prefix='{sample}',
         db=directory(os.path.join(DB_DIR_PATH,"eggnog_" + VERSION_EGGNOG)),
         sensmode=config.get("eggnog_sensmode")
     threads:
@@ -1075,7 +1076,7 @@ rule run_eggnog:
         """
         mkdir -p {output.outdir}/tmp
         emapper.py --cpu {threads} -i {input.protein} --itype proteins -m diamond --sensmode {params.sensmode} \
-          --dbmem --output eggnog --output_dir {output.outdir} --temp_dir {output.outdir}/tmp \
+          --dbmem --output eggnog --output_dir {output.outdir} --temp_dir {output.outdir}/tmp --output {params.prefix}  \
           --data_dir {params.db} --override > {log} 2>&1
         rm -r {output.outdir}/tmp
         """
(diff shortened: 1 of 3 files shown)


LeeBergstrand:
Passes `--dry-run` but is still testing with real data.

rotary/rules/rotary.smk
```diff
@@ -314,17 +314,17 @@ rule qc_long_length_stats:
...
 
 
 rule assembly_flye:
     input:
-        "{sample}/qc_long/nanopore_qc.fastq.gz"
+        "{sample}/qc_long/{sample}_nanopore_qc.fastq.gz"
     output:
-        assembly="{sample}/assembly/flye/assembly.fasta",
-        info="{sample}/assembly/flye/assembly_info.txt",
+        assembly="{sample}/assembly/flye/{sample}_assembly.fasta",
```
jmtsuji, line 326: Flye does not allow you to specify the names of the output files in its output folder. So perhaps we should leave `assembly.fasta` and `assembly_info.txt` as-is? Otherwise, we will need to add `mv` commands to the shell section here.

jmtsuji: (This comment also applied to line 327 below)

rotary/rules/rotary.smk
```diff
@@ -350,12 +350,12 @@ rule assembly_flye:
 # TODO - the math for memory per thread should really be done somewhere other than 'resources' - what is best practice?
 rule assembly_end_repair:
     input:
-        qc_long_reads="{sample}/qc_long/nanopore_qc.fastq.gz",
-        assembly="{sample}/assembly/flye/assembly.fasta",
-        info="{sample}/assembly/flye/assembly_info.txt"
+        qc_long_reads="{sample}/qc_long/{sample}_nanopore_qc.fastq.gz",
+        assembly="{sample}/assembly/flye/{sample}_assembly.fasta",
```
jmtsuji, line 354: See my comment in rule assembly_flye -- the assembly and assembly_info files might need to have the sample prefix removed

rotary/rules/rotary.smk
```diff
@@ -350,12 +350,12 @@ rule assembly_flye:
...
+        info="{sample}/assembly/flye/{sample}_assembly_info.txt"
     output:
-        assembly="{sample}/assembly/end_repair/repaired.fasta",
-        info="{sample}/assembly/end_repair/repaired_info.tsv",
+        assembly="{sample}/assembly/end_repair/{sample}_repaired.fasta",
```
jmtsuji, line 357: Like Flye, my code also doesn't allow you to specify the files names in the output folder. So maybe best to keep these as `repaired.fasta` and `repaired_info.tsv`? They will be symlinked below anyway, so the name could be adjusted then.

rotary/rules/rotary.smk
```diff
@@ -396,11 +396,11 @@ rule assembly_end_repair:
 
 rule finalize_assembly:
     input:
-        assembly="{sample}/assembly/end_repair/repaired.fasta",
-        info="{sample}/assembly/end_repair/repaired_info.tsv"
+        assembly="{sample}/assembly/end_repair/{sample}_repaired.fasta",
```
jmtsuji, line 399: These two input file names might need to have the prefix removed, based on the discussion I left rule assembly_end_repair. But the output files here (which will be the summary files that the user will interact with) are OK to keep the sample prefix.

rotary/rules/rotary.smk
```diff
@@ -514,13 +514,13 @@ rule polish_polypolish:
...
-        qc_short_r2 = "{sample}/{sample}_R2.fastq.gz",
-        polished = "{sample}/polish/polypolish/polypolish.fasta"
+        qc_short_r1 = "{sample}/raw/{sample}_R1.fastq.gz",
+        qc_short_r2 = "{sample}/raw/{sample}_R2.fastq.gz",
+        polished = "{sample}/polish/polypolish/{sample}_polypolish.fasta"
     output:
-        polca_output = "{sample}/polish/polca/polca.fasta",
-        polypolish_sam = temp("{sample}/polish/polca/polypolish.fasta.unSorted.sam"),
-        polypolish_bam = temp("{sample}/polish/polca/polypolish.fasta.alignSorted.bam")
+        polca_output = "{sample}/polish/polca/{sample}_polca.fasta",
```
jmtsuji, line 521: Line 540 (symlink) will need to be changed to match this name change. I think I didn't use `{output.polca_output}` there because of the odd `cd` steps. However, I think line 540 should be changed to properly use `{output.polca_output}` instead of hard-coding the final name.

rotary/rules/rotary.smk
```diff
@@ -1055,7 +1055,7 @@ rule run_dfast:
 # TODO - add option to control whether --dbmem flag is set (uses more RAM but does faster analysis)
 rule run_eggnog:
     input:
-        protein="{sample}/annotation/dfast/protein.faa",
+        protein="{sample}/annotation/dfast/{sample}_protein.faa",
         install_finished=os.path.join(DB_DIR_PATH,"checkpoints","eggnog_" + VERSION_EGGNOG)
     output:
         eggnog_annotations="{sample}/annotation/eggnog/eggnog.emapper.annotations",
```
jmtsuji, line 1061: Will adding `--output [sample_id]` change the name of this output file, e.g., to `[sample_id].emapper.annotations`?

jmtsuji (changes requested):
Thanks for working on these naming and pathing changes! I have flagged a few possible naming issues, but otherwise it looks good. I might end up cleaning up some of the issues I flagged myself before running an end-to-end test on this branch.

rotary/rules/rotary.smk
LeeBergstrand: @jmtsuji Thanks! Added mv commands as part of my debugging efforts.

rotary/rules/rotary.smk
LeeBergstrand: @jmtsuji Added mv commands to address this issue during my debugging.

rotary/rules/rotary.smk
LeeBergstrand: Seems to run fine I think they pick up the sample names from prior rules.

rotary/rules/rotary.smk
LeeBergstrand: I will check is `ln` allows longer paths.

rotary/rules/rotary.smk
LeeBergstrand: @jmtsuji The output flag adds a prefix to all the annotation output files.

``` 
--output FILE_PREFIX, -o FILE_PREFIX
                        base name for output files
```

LeeBergstrand: I will see if it causes the `emapper.annotations` to fail.

LeeBergstrand: So `emapper.annotations` now has to be proceeded by `{sample}`

LeeBergstrand: Solved.

rotary/rules/rotary.smk
LeeBergstrand: I think its more straightforward to add wildcards to this since there is a `cd`. Added.

jmtsuji (approved):
Nice work -- this looks good to me. I'm working off-site and don't have access to the server for end-to-end testing, but shall we go ahead and merge this into add-multi-sample and then do end-to-end testing there?

**Update**: there were no merge conflicts, so I created a PR of my short read QC work into this branch at #75 . The comment below is no longer relevant:
_One more thought -- I made a new branch from this branch yesterday for short read QC work. What is the best way to avoid merge conflicts between my short read QC branch and the other branches down the road, given that some additional work has been done on this multi sample pathing branch? Would it be best to merge the last ~24 hour's changes from this branch into my short read QC branch and deal with any merge conflicts there, for example, rather than wait to deal with merge conflicts later? This would let the short read QC branch merge directly into add-muti-sample or into develop later on without much trouble. Thanks!_

LeeBergstrand:
> Nice work -- this looks good to me. I'm working off-site and don't have access to the server for end-to-end testing, but shall we go ahead and merge this into add-multi-sample and then do end-to-end testing there?
> 
> **Update**: there were no merge conflicts, so I created a PR of my short read QC work into this branch at #75 . The comment below is no longer relevant: _One more thought -- I made a new branch from this branch yesterday for short read QC work. What is the best way to avoid merge conflicts between my short read QC branch and the other branches down the road, given that some additional work has been done on this multi sample pathing branch? Would it be best to merge the last ~24 hours' changes from this branch into my short read QC branch and deal with any merge conflicts there, for example, rather than wait to deal with merge conflicts later? This would let the short read QC branch merge directly into add-muti-sample or develop later on without much trouble. Thanks!_

So, the general best practice is to push and merge as often as possible so your branches stay similar. However, there are techniques like git rebase (https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase) that can be used to pluck branches and move their start points around. Rebases can only be used locally as they rewrite the history.

LeeBergstrand:
> > Nice work -- this looks good to me. I'm working off-site and don't have access to the server for end-to-end testing, but shall we go ahead and merge this into add-multi-sample and then do end-to-end testing there?
> > **Update**: there were no merge conflicts, so I created a PR of my short read QC work into this branch at #75 . The comment below is no longer relevant: _One more thought -- I made a new branch from this branch yesterday for short read QC work. What is the best way to avoid merge conflicts between my short read QC branch and the other branches down the road, given that some additional work has been done on this multi sample pathing branch? Would it be best to merge the last ~24 hours' changes from this branch into my short read QC branch and deal with any merge conflicts there, for example, rather than wait to deal with merge conflicts later? This would let the short read QC branch merge directly into add-muti-sample or develop later on without much trouble. Thanks!_
> 
> So, the general best practice is to push and merge as often as possible so your branches stay similar. However, there are techniques like git rebase (https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase) that can be used to pluck branches and move their start points around. Rebases can only be used locally as they rewrite the history.

Also, sometimes pull the parent branch into my feature branch everyone and a while to stay updated.

jmtsuji:
@LeeBergstrand Got it -- helpful tips! Thank you for merging changes from develop into the short read QC branch in #77

Issue #38, opened 2023-10-16
Add automated Illumina QC (similar to ATLAS).


LeeBergstrand:
I don't think we can directly port over the ATLAS QC directly over. However, we can at least copy some of the rules and modify them to be simpler in terms of variables called.

jmtsuji:
OK, seems reasonable to make our own basic QC workflow. I guess the basic QC workflow would include the following (inspired by ATLAS)?
- Quality trimming
- Adapter removal
- Contamination (e.g., PhiX) removal??

I think the following steps are also performed by ATLAS, but they might not be needed in our case (because genome data is simpler than metagenome data, and we won't perform de novo assembly of short reads):
- Read deduplication
- Sorting of short reads to make assembly more efficient

Lastly, I don't think we need to perform error correction of short reads or paired-end read merging, given that the short reads will only be used for polishing the long read-derived assembly.

How does that sound?

LeeBergstrand:
@jmtsuji I agree with your assessment. We may also apply the decontamination to the long reads for decontamination.  

In Atlas you can actually specify files to filter by, for example human genomes: https://github.com/metagenome-atlas/atlas/blob/2871b8a1d6093b4caf1cad03145bcd3daa001f71/docs/advanced/qc.rst

LeeBergstrand:
Host contamination is becoming. So many database genomes have host human genes in them that it is starting to ruin some short-read taxonomy assignment algorithms. They start classifying host genes as bacterial because there's a close match to human genes found in contaminated reference genomes.

https://twitter.com/profbootyphd/status/1688386329405505536?s=61&t=dJHfNw7Lsr9K6D3WRAk6qA

In response to: https://www.nature.com/articles/s41586-020-2095-1

jmtsuji:
Sounds good to allow for multiple references for decontamination. Yeah, I've also heard about the human DNA contamination issue... very bad for short read taxonomy algorithms indeed...

(As a side note, one positive thing about rotary is that the majority of the output genomes will likely be complete, which minimizes the risks of human DNA contamination)

Not sure of the best tool for long read decontamination -- the high errors rates in the raw long reads could be an issue. If you find a tool for this that looks reasonable, then yes, feel free to add it to the workflow.

[...]

PR #75, opened 2023-11-08
Short read QC

jmtsuji:
This code adds basic short read QC functionality to rotary to address #38. Summary of changes:
- A new master checkpoint for QC was created in `rule qc`. This selects (on a per-run basis, for the MVP version) whether short + long read QC will be performed or just long read QC
- Short read QC starts with tossing out junk reads (e.g., with ambiguous bases) and checking file integrity with `rule short_read_reformat`
- Adapter removal (from the right side of reads) and quality trimming are then performed in the same step, `rule short_read_adapter_and_quality_trimming`. Quality trimming is performed after adapter removal so that adapters near the ends of reads can be more accurately identified (e.g., before some bases are clipped off the ends). I think ATLAS takes the same approach.
  - TODO: the adapters file from ATLAS is directly downloaded from Zenodo - do we want to make and host our own file?
  - Note that adapter trimming and quality trimming are currently non-optional. We can add the ability to toggle these on/off if you would like.
  - I set a relatively aggressive q-trim threshold of Q20
- Contaminant filtration is the last step, via `rule short_read_contamination_filter`. This step is optional based on a variable in the config.
  - I built in PhiX genome and human genome contaminant filtration into rotary. These genomes will be auto-downloaded into the DB folder if the user requests to screen with them.
  - The user can optionally add a single additional FastA nucleotide file as a screening reference.
  
I did not add good logging for QC stats. The user needs to manually look at log files to understand what happened during QC. In future, it could be nice to add a rule that parses the log files to summarize QC results.

Lastly, the current code passes `--dry-run` on my machine but has only been tested in smaller chunks with real data. A proper end-to-end test will be needed at some point (maybe once everything is merged into #71 ?).

@LeeBergstrand Mind adding your review? It looks like there will be no merge conflicts with improve_multi_sample_pathing (the branch that this branch was derived from), which is good news! Thanks.

### DIFF
--- a/rotary/config.yaml
+++ b/rotary/config.yaml
@@ -100,3 +100,5 @@ gtdbtk_mode: "default"
 # EggNOG annotation sensitivity mode (using DIAMOND)
 #    Options: fast, mid-sensitive, sensitive, more-sensitive, very-sensitive, ultra-sensitive
 eggnog_sensmode: "sensitive"
+# The search tool that emapper.py should use (diamond or mmseqs).
+eggnog_search_tool: 'mmseqs'
--- a/rotary/config.yaml
+++ b/rotary/config.yaml
@@ -30,9 +30,10 @@ minimum_average_quality_score_post_trim: 20
 minimum_read_length: 10
 # K-mer length for filtering contaminant genome info out of the data
 contamination_filter_kmer_length: 27
-# Contaminant reference genomes to use for the contamination search
-# If you want to provide your own custom genome(s), add the file path(s) to the list below (for FastA nucleotide files)
+# Contaminant reference genomes (provided within rotary) for filtering contaminants out of raw short reads
 contamination_references: ["PhiX", "human"]
+# If you want to provide a custom genome for contaminant filtration, add the path to the FastA nucleotide file here:
+custom_contamination_reference: "None"
 
 ## Assembly
 # nano-hq (if <5% error) or nano-raw
--- a/rotary/config.yaml
+++ b/rotary/config.yaml
@@ -15,6 +15,25 @@ minlength: 1000
 # Minimum average quality (PHRED) score. Q13 is ~5% expected error rate.
 minavgquality: 13
 
+## Short read QC
+# K-mer length for adapter trimming (this k-mer length should be shorter than the adapters)
+adapter_trimming_kmer_length: 27
+# Minimum adapter length to be detectable on the (right-hand) end of a read. Lower number = more false positives
+minimum_detectable_adapter_length_on_read_end: 6
+# Whether to perform quality trimming on the "right" side of each read or on "both" the left and right sides
+quality_trim_direction: "both"
+# Quality score cutoff for quality trimming
+quality_trim_score_cutoff: 20
+# Minimum average quality score required for a read after quality trimming
+minimum_average_quality_score_post_trim: 20
+# Minimum read length after quality trimming (otherwise is discarded)
+minimum_read_length: 10
+# K-mer length for filtering contaminant genome info out of the data
+contamination_filter_kmer_length: 27
+# Contaminant reference genomes to use for the contamination search
+# If you want to provide your own custom genome(s), add the file path(s) to the list below (for FastA nucleotide files)
+contamination_references: ["PhiX", "human"]
+
 ## Assembly
 # nano-hq (if <5% error) or nano-raw
 flye_input_mode: "nano-hq"
--- a/rotary/config.yaml
+++ b/rotary/config.yaml
@@ -42,13 +42,15 @@ minimum_read_length_quality_trim: 10
 # Search for and remove reads that match contaminant reference sequences?
 perform_contaminant_filtration: "True"
 # K-mer length for filtering contaminants
-contamination_filter_kmer_length: 27
-# Contaminant reference genomes (provided within rotary) for filtering contaminants out of raw short reads
-#  If you do not want to filter with either of these (but do want to filter with your own custom genome below), then
-#  set contamination_references to []
-contamination_references: ["PhiX", "human"]
-# If you want to provide a custom genome for contaminant filtration, add the path to the FastA nucleotide file here:
-custom_contamination_reference: "False"
+contamination_filter_kmer_length: 31
+# NCBI genome assembly accessions to use for filtering contaminants from short read data.
+#   Default accessions: GCF_000819615.1 (PhiX), GCF_000001405.40 (human)
+#   NOTE that contaminant filtration for the human genome takes ~60 GB RAM.
+#   If you do not want to use any NCBI genome accessions for filtering, then set this variable to []
+contamination_references_ncbi_accessions: ["GCF_000819615.1", "GCF_000001405.40"]
+# Filepaths to FastA nucleotide files to use for filtering contaminants from short read data.
+#   If you do not want to use any custom FastA files for filtering, then set this variable to []
+contamination_references_custom_filepaths: []
 
 ## Assembly
 # nano-hq (if <5% error) or nano-raw
--- a/rotary/config.yaml
+++ b/rotary/config.yaml
@@ -16,25 +16,36 @@ minlength: 1000
 minavgquality: 13
 
 ## Short read QC
+## Adapter trimming
+# Perform adapter trimming from the right side of reads? Identifies adapters using both a built-in database.
+perform_adapter_trimming: "True"
 # K-mer length for adapter trimming (this k-mer length should be shorter than the adapters)
 adapter_trimming_kmer_length: 27
 # Minimum adapter length to be detectable on the (right-hand) end of a read. Lower number = more false positives
 minimum_detectable_adapter_length_on_read_end: 6
+# In addition to database-based trimming, trim the overhang of reads if they completely overlap
+#  (Requires at least a 40 bp insert and 14 bp read overlap)
+overlap_based_trimming: "True"
+# Minimum read length after adapter trimming (otherwise is discarded)
+minimum_read_length_adapter_trim: 10
+## Quality trimming
 # Whether to perform quality trimming on the right ("r") side of each read or on both the left and right ("rl") sides
+#  Otherwise, specify "f" to skip quality trimming altogether.
 quality_trim_direction: "rl"
 # Quality score cutoff for quality trimming
 quality_trim_score_cutoff: 20
-# Minimum average quality score required for a read after quality trimming
+# Minimum average quality score required for a read after quality trimming (disabled if quality_trim_direction is "f")
 minimum_average_quality_score_post_trim: 20
 # Minimum read length after quality trimming (otherwise is discarded)
-minimum_read_length: 10
+minimum_read_length_quality_trim: 10
+## Contaminant filtration
 # Search for and remove reads that match contaminant reference sequences?
 perform_contaminant_filtration: "True"
 # K-mer length for filtering contaminants
 contamination_filter_kmer_length: 27
 # Contaminant reference genomes (provided within rotary) for filtering contaminants out of raw short reads
-# If you do not want to filter with either of these (but do want to filter with your own custom genome below), then
-# set contamination_references to []
+#  If you do not want to filter with either of these (but do want to filter with your own custom genome below), then
+#  set contamination_references to []
 contamination_references: ["PhiX", "human"]
 # If you want to provide a custom genome for contaminant filtration, add the path to the FastA nucleotide file here:
 custom_contamination_reference: "False"
--- a/rotary/config.yaml
+++ b/rotary/config.yaml
@@ -35,7 +35,7 @@ contamination_filter_kmer_length: 27
 # Contaminant reference genomes (provided within rotary) for filtering contaminants out of raw short reads
 contamination_references: ["PhiX", "human"]
 # If you want to provide a custom genome for contaminant filtration, add the path to the FastA nucleotide file here:
-custom_contamination_reference: "None"
+custom_contamination_reference: "False"
 
 ## Assembly
 # nano-hq (if <5% error) or nano-raw
--- a/rotary/config.yaml
+++ b/rotary/config.yaml
@@ -33,6 +33,8 @@ perform_contaminant_filtration: "True"
 # K-mer length for filtering contaminants
 contamination_filter_kmer_length: 27
 # Contaminant reference genomes (provided within rotary) for filtering contaminants out of raw short reads
+# If you do not want to filter with either of these (but do want to filter with your own custom genome below), then
+# set contamination_references to []
 contamination_references: ["PhiX", "human"]
 # If you want to provide a custom genome for contaminant filtration, add the path to the FastA nucleotide file here:
 custom_contamination_reference: "False"
--- a/rotary/config.yaml
+++ b/rotary/config.yaml
@@ -28,7 +28,9 @@ quality_trim_score_cutoff: 20
 minimum_average_quality_score_post_trim: 20
 # Minimum read length after quality trimming (otherwise is discarded)
 minimum_read_length: 10
-# K-mer length for filtering contaminant genome info out of the data
+# Search for and remove reads that match contaminant reference sequences?
+perform_contaminant_filtration: "True"
+# K-mer length for filtering contaminants
 contamination_filter_kmer_length: 27
 # Contaminant reference genomes (provided within rotary) for filtering contaminants out of raw short reads
 contamination_references: ["PhiX", "human"]
--- a/rotary/config.yaml
+++ b/rotary/config.yaml
@@ -16,6 +16,7 @@ minlength: 1000
 minavgquality: 13
 
 ## Short read QC
+
 ## Adapter trimming
 # Perform adapter trimming from the right side of reads? Identifies adapters using both a built-in database.
 perform_adapter_trimming: "True"
@@ -28,6 +29,7 @@ minimum_detectable_adapter_length_on_read_end: 6
 overlap_based_trimming: "True"
 # Minimum read length after adapter trimming (otherwise is discarded)
 minimum_read_length_adapter_trim: 10
+
 ## Quality trimming
 # Whether to perform quality trimming on the right ("r") side of each read or on both the left and right ("rl") sides
 #  Otherwise, specify "f" to skip quality trimming altogether.
@@ -38,9 +40,8 @@ quality_trim_score_cutoff: 20
 minimum_average_quality_score_post_trim: 20
 # Minimum read length after quality trimming (otherwise is discarded)
 minimum_read_length_quality_trim: 10
+
 ## Contaminant filtration
-# Search for and remove reads that match contaminant reference sequences?
-perform_contaminant_filtration: "True"
 # K-mer length for filtering contaminants
 contamination_filter_kmer_length: 31
 # NCBI genome assembly accessions to use for filtering contaminants from short read data.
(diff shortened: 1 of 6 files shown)


rotary/rules/rotary.smk
```diff
@@ -8,9 +8,11 @@ import itertools
 from snakemake.utils import min_version
 
 from rotary.sample import parse_sample_tsv
-from rotary.utils import symlink_or_compress
+from rotary.utils import symlink_or_compress, get_contamination_reference_files
 
 
+PHIX_GENOME_ACCESSION = "GCF_000819615.1"
```
LeeBergstrand, line 14: @jmtsuji I think we should eventually move these versions to a separate configuration file. The database version list is starting to get quite long.

rotary/rules/rotary.smk
```diff
@@ -36,13 +38,66 @@ else:
...
+    shell:
+        """        
+        echo "### Downloading genome: {params.name} {params.accession} ###" > {log}
+        datasets download genome accession {params.accession} --include genome --filename {output.zip} 2>> {log}
+        unzip -d {output.zip_dir} {output.zip} > /dev/null
+        
+        # Confirm that there is only one genome file matching the expected pattern in the unzipped folder
+        genome_file=($(find {output.zip_dir}/ncbi_dataset/data/{params.accession} -type f -name "{params.accession}_*_genomic.fna"))
+        if [[ "${{#genome_file[@]}}" == 1 ]]; then
+          gzip -c "${{genome_file[0]}}" > {output.genome}
```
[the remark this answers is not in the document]
LeeBergstrand: If it's like 1-3 mb genome it's probably okay. I assume it won't take long.

rotary/rules/rotary.smk
```diff
@@ -312,16 +367,180 @@ rule qc_long_length_stats:
...
+        expand("{sample}/qc/{sample}_QC_long.fastq.gz", sample=SAMPLE_NAMES),
+        expand("{sample}/stats/{sample}_qc_long_length_stats.txt", sample=SAMPLE_NAMES)
     output:
         temp(touch("checkpoints/qc_long"))
 
 
+rule short_read_reformat:
+    """
+    Makes the input format of the short reads consistent (e.g., by trimming read descriptions and tossing invalid 
+    nucleotide characters). Threads are locked at a maximum of 4 because this code is IO limited.
```
LeeBergstrand, line 391: @jmtsuji I think snakemake has a way to override resources for specific rules. What is interesting about this is that if you put this code on a RAM disk or potentially SSD you can use more threads. I think locking it at 4 threads is good for now.

LeeBergstrand: I found that when I was running HMMER3 on ZFS data with the RAM cache I could use more threads than the 8 it normally bottlenecks at because the RAM cache was able to feed the algorithm fast enough to make a difference.

rotary/rules/rotary.smk
```diff
@@ -312,16 +367,180 @@ rule qc_long_length_stats:
         length_stats.to_csv(output[0], sep='\t', header=None, index=True)
 
 
+rule finalize_qc_long:
+    input:
+        "{sample}/qc/long/{sample}_nanopore_qc.fastq.gz"
+    output:
+        "{sample}/qc/{sample}_QC_long.fastq.gz"
```
LeeBergstrand, line 374: @jmtsuji This might be a bit nitpicky, but I was wondering if there was a reason why QC is capitalized? All the other files with QC in their file names have the 'qc' lower case.

rotary/utils.py
```diff
@@ -100,3 +100,39 @@ def symlink_or_compress(in_file_path, out_file_path):
...
+    :param db_path: path to the rotary database folder
+    :param phix_accession: NCBI assembly accession code for the PhiX genome assembly
+    :param human_accession: NCBI assembly accession code for the human genome assembly
+    :return: list of file paths for the contamination reference sequences
+    """
+
+    contamination_reference_paths = []
+
+    if isinstance(config_entry, list) is True:
+
```
LeeBergstrand, line 119: @jmtsuji I haven't brought this up before, but in some situations, you can drop the blank lines below the for and if statements and let the indentation handle the readability.

```python
if isinstance(config_entry, list) is True:
        for contamination_entry in config_entry:
            if contamination_entry.lower() == 'phix':
                contamination_reference_paths.append(os.path.join(db_path, 'contamination_references',
                                                                  f'PhiX__{phix_accession}.fna.gz'))
            elif contamination_entry.lower() == 'human':
                contamination_reference_paths.append(os.path.join(db_path, 'contamination_references',
                                                                  f'human__{human_accession}.fna.gz'))
            else:
                raise ValueError(f'Contaminant genome must be PhiX or human, but you provided {contamination_entry}')
else:
    raise TypeError(f'Contaminant genome names must be in a list, but you provided type {type(config_entry)}. '
                    'Try editing the config file and make sure that contamination_references is a list like '
                    '[PhiX] or [PhiX, human], not just "PhiX" or "human".')
```

rotary/utils.py
```diff
@@ -100,3 +100,39 @@ def symlink_or_compress(in_file_path, out_file_path):
...
+    :param config_entry: value of config.get('contamination_references'); should be a list
+    :param db_path: path to the rotary database folder
+    :param phix_accession: NCBI assembly accession code for the PhiX genome assembly
+    :param human_accession: NCBI assembly accession code for the human genome assembly
+    :return: list of file paths for the contamination reference sequences
+    """
+
+    contamination_reference_paths = []
+
+    if isinstance(config_entry, list) is True:
```
LeeBergstrand, line 118: Also, you can remove the is True here and it should still work.:

```python
isinstance(config_entry, list):
   do stuff..
```

I usually do this with functions that evaluate something for example functions that evaluate only true or false. I.e., is the config entry an instance of list? Yes or No.

rotary/utils.py
LeeBergstrand: I usually use blank lines to break out logical blocks. For example:

```python
if x: # no space here because it's redundant because of the indent.
   data=some_func()
           # space here to break out the for loop away from the other code. Since blam is filled
   blam=[] # by the loop, we put it directly above the loop with no space and break it out.  
   for datum in data:
       if datum.cool == True: # Some libraries like numpy they have their own boolean values  
          blam.append(datum)  # (```np.bool```) so you have to use == to force an evaluation. 

   if len(blam) > 1: # Blank line between for loop and if statement since they are different logics.
       print(blam)
```

LeeBergstrand: @jmtsuji Just some examples of how I use spacing.

LeeBergstrand:
@jmtsuji This looks good so far. I left a few comments.

rotary/rules/rotary.smk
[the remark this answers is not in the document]
jmtsuji: The issue is the human genome -- it's ~1-3 GB (I was surprised by the size!). I tested it on my Mac, and the compression time with standard gzip is about 5-10 minutes, which is tolerable but a bit long. I've added pigz now to the rule (with threads capped at 4 max like the short_read_reformat rule below).

rotary/rules/rotary.smk
jmtsuji: I agree -- I was thinking the same thing. Will convert to issue.

jmtsuji: Done -- #78

rotary/rules/rotary.smk
jmtsuji: Good point that some hardware configurations could support higher thread counts. One good thing is that snakemake should be able to run multiple jobs in parallel that are below the total resources allotted to snakemake. So even if capping this job at 4 threads does not maximize the thread usage in all setups, at least other rules should be able to run simultaneously to fill the space. I agree, let's keep it at 4 threads for now, but we can always change this in future (or add another config variable somewhere, whether exposed or hidden, for the thread count of small thread jobs).

rotary/rules/rotary.smk
jmtsuji: Thanks for noticing this! I've converted all to lowercase.

rotary/utils.py
jmtsuji: Good to know I can simplify this -- I was trying to keep the True/False test obvious in the code for readability, but I think you're right that the function name itself makes it pretty clear what the code is doing. Just addressed this.

jmtsuji:
@LeeBergstrand Addressed your comments -- mind taking another look?

Also, what are your thoughts on the following (pasted from the PR description above)?
> - TODO: the adapters file from ATLAS is directly downloaded from Zenodo - do we want to make and host our own file?
> - Note that adapter trimming and quality trimming are currently non-optional. We can add the ability to toggle these on/off if you would like.
> - I set a relatively aggressive q-trim threshold of Q20 [as default]

LeeBergstrand:
@jmtsuji 

> > * TODO: the adapters file from ATLAS is directly downloaded from Zenodo - do we want to make and host our own file?

We could put this down as a long-term TODO. I wonder if the archive will go down or if ATLAS will stop using this URL.

> > * Note that adapter trimming and quality trimming are currently non-optional. We can add the ability to toggle these on/off if you would like.

This should be implemented. However, I would like to merge the QC code into develop beforehand.

> > * I set a relatively aggressive q-trim threshold of Q20 [as default]

This is fine for Illumina, which should have reasonable error rates. We should have more stringent cutoffs since we are polishing with the short reads.

rotary/rules/rotary.smk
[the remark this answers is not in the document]
LeeBergstrand: @jmtsuji My mistake; I forgot that this was the human genome, I was underreading. Yes, it makes a lot of sense to use pigz in this situation.

jmtsuji:
@LeeBergstrand 
> We could put this down as a long-term TODO. I wonder if the archive will go down or if ATLAS will stop using this URL.

Addressed - the code now pulls from a rotary-specific data repo that contains the adapters file: https://zenodo.org/records/10087395
Attribution is given to the ATLAS repo.

jmtsuji:
> This should be implemented. However, I would like to merge the QC code into develop beforehand.

OK, sounds good.

> This is fine for Illumina, which should have reasonable error rates. We should have more stringent cutoffs since we are polishing with the short reads.

Great, let's proceed with Q20 for now, then.

jmtsuji:
> > Note that adapter trimming and quality trimming are currently non-optional. We can add the ability to toggle these on/off if you would like.

> This should be implemented. However, I would like to merge the QC code into develop beforehand.

@LeeBergstrand FYI, I had a bit of extra time, so I went ahead and added the ability to toggle adapter and quality trimming on/off into the code. Basically:
- I had to split adapter trimming and quality trimming into 2 separate rules
- The adapter trimming rule does not run if the config flag `perform_adapter_trimming` is set to False
- The quality trimming rule always runs, but if `quality_trim_direction` is set to "f" in the config (an actual setting in bbduk), then the reads are just passed through the rule without actual trimming. This simplified the logic in the snakefile quite a bit.

rotary/rules/rotary.smk
```diff
@@ -312,16 +369,228 @@ rule qc_long_length_stats:
...
+    conda:
+        "../envs/qc.yaml"
+    log:
+        "{sample}/logs/qc/short/short_read_contamination_filter.log"
+    benchmark:
+        "{sample}/benchmarks/qc/short/short_read_contamination_filter.benchmark.txt"
+    params:
+        contamination_filter_kmer_length = config.get("contamination_filter_kmer_length"),
+        all_contaminant_references = ','.join(get_contamination_reference_files(config.get("contamination_references"),
+                DB_DIR_PATH, PHIX_GENOME_ACCESSION, HUMAN_GENOME_ACCESSION)) \
```
LeeBergstrand, line 525: @jmtsuji Move this to a function passed to params (I think that's possible?), or use this code to assign a global variable at the top of the file. This is too long of an inline if-else statement. Once the code is moved above, you can convert it to a traditional if-else statement using pycharm (right-click the if and select show context actions ). Also, you could assign the result of `get_contamination_reference_files(config.get("contamination_references"), DB_DIR_PATH, PHIX_GENOME_ACCESSION, HUMAN_GENOME_ACCESSION)` to the second global variable, which could be used in the prior mentioned if-else statement and could also be assigned to ```input.contamination_references```  for rule ```short_read_contamination_filter```

rotary/rules/rotary.smk
```diff
@@ -312,16 +369,228 @@ rule qc_long_length_stats:
...
+    Note that contamination_references in the input is only there to trigger download of the genome files; the actual 
+    text to be input into the shell (with proper comma separation) is prepared in all_contaminant_references in params. 
+    The function get_contamination_reference_files called by this rule returns genome paths that can be understood by 
+    rule download_contamination_reference, which allows the download rule to be triggered as needed depending on which 
+    genomes the user wants to use as references. 
+    """
+    input:
+        short_r1 = "{sample}/qc/short/{sample}_quality_trim_R1.fastq.gz",
+        short_r2 = "{sample}/qc/short/{sample}_quality_trim_R2.fastq.gz",
+        contamination_references = get_contamination_reference_files(config.get("contamination_references"),
```
LeeBergstrand, line 509: move to global variable.

rotary/utils.py
```diff
@@ -100,3 +100,34 @@ def symlink_or_compress(in_file_path, out_file_path):
         os.symlink(in_file_path, out_file_path)
     else:
         gzip_file(in_file_path, out_file_path)
+
+
+def get_contamination_reference_files(config_entry: list, db_path: str, phix_accession: str, human_accession: str):
```
LeeBergstrand, line 105: Change the name of this function to `get_contamination_reference_file_paths.` Generally, when I put the word `file` of a function, I expect it to return a file handle, which is the object created by the `open('file/path).` This function returns the paths to the files, not the files or handles themselves.

LeeBergstrand (changes requested):

LeeBergstrand:
Improve by putting contaminant NCBI IDs in the config and also a list of paths to custom contaminant FASTA. As discussed below:

image

rotary/rules/rotary.smk
jmtsuji: @LeeBergstrand This variable was simplified so that an if-else statement is no longer needed. It's still a bit long, though (mostly because of long variable and function names) -- what do you think?

rotary/rules/rotary.smk
jmtsuji: Done

rotary/utils.py
jmtsuji: Done - good point!

jmtsuji:
> Improve by putting contaminant NCBI IDs in the config and also a list of paths to custom contaminant FASTA.

Done! Let me know your thoughts on the updated code.

Aside from addressing the core changes you requested, some other things were also added:
- warning was added in the config file about ~60 GB memory needed for human genome-based decontamination
- pigz / unpigz is also forced on now

rotary/rules/rotary.smk
jmtsuji: Also, note that this will fail with a somewhat cryptic error message if the user provides bad input (e.g., a string, not a list) to `contamination_references_custom_filepaths` in the config. We could add a sanity check for this in the config check proposed in #79 . Otherwise, I could make a function to validate these input files.

rotary/config.yaml
```diff
@@ -15,6 +15,43 @@ minlength: 1000
...
+quality_trim_direction: "rl"
+# Quality score cutoff for quality trimming
+quality_trim_score_cutoff: 20
+# Minimum average quality score required for a read after quality trimming (disabled if quality_trim_direction is "f")
+minimum_average_quality_score_post_trim: 20
+# Minimum read length after quality trimming (otherwise is discarded)
+minimum_read_length_quality_trim: 10
+## Contaminant filtration
+# Search for and remove reads that match contaminant reference sequences?
+perform_contaminant_filtration: "True"
```
LeeBergstrand, line 43: @jmtsuji What are your thoughts about removing this parameter and having it so that the sets both `contamination_references_ncbi_accessions` and `contamination_references_custom_filepaths` to `[]` to turn off the contamination filtering. Having this `perform_contaminant_filtration` flag is redundant now with your recent changes. What are your thoughts?

rotary/rules/rotary.smk
```diff
@@ -312,16 +369,222 @@ rule qc_long_length_stats:
...
+rule short_read_contamination_filter:
+    """
+    Filters short reads based on match to a reference. 
+    Note that contamination_references in the input is only there to trigger download of the genome files; the actual 
+    text to be input into the shell (with proper comma separation) is prepared in all_contaminant_references in params.
+    """
+    input:
+        short_r1 = "{sample}/qc/short/{sample}_quality_trim_R1.fastq.gz",
+        short_r2 = "{sample}/qc/short/{sample}_quality_trim_R2.fastq.gz",
+        contamination_references = CONTAMINANT_REFERENCE_GENOMES_NCBI
```
LeeBergstrand, line 507: @jmtsuji Will this fail if CONTAMINANT_REFERENCE_GENOMES_NCBI = None? How will this work if there are no NCBI contamination references?

rotary/rules/rotary.smk
```diff
@@ -312,16 +369,222 @@ rule qc_long_length_stats:
...
+    shell:
+        """
+        bbduk.sh -Xmx{resources.mem}g threads={threads} in={input.short_r1} in2={input.short_r2} \
+          out={output.short_r1} out2={output.short_r2} qhist={output.quality_histogram} \
+          qtrim={params.quality_trim_direction} trimq={params.quality_trim_cutoff} minlength={params.min_read_length} \
+          minavgquality={params.min_average_quality} overwrite=t interleaved=f qin=33 pigz=t unpigz=t \
+          2> {log}
+        """
+
+
```
LeeBergstrand, line 497: Shouldn't this be in an if block?

rotary/rules/rotary.smk
```diff
@@ -312,16 +369,222 @@ rule qc_long_length_stats:
...
+        filter_stats= "{sample}/stats/{sample}_short_read_contamination_filter.txt"
+    conda:
+        "../envs/qc.yaml"
+    log:
+        "{sample}/logs/qc/short/short_read_contamination_filter.log"
+    benchmark:
+        "{sample}/benchmarks/qc/short/short_read_contamination_filter.benchmark.txt"
+    params:
+        contamination_filter_kmer_length = config.get("contamination_filter_kmer_length"),
+        all_contaminant_references = ','.join(CONTAMINANT_REFERENCE_GENOMES_NCBI +
```
LeeBergstrand, line 521: I would move this logic to the top and assign it to a variable called "ALL_CONTAMINANT_REFERENCES." If contaminant references are none, then skip the contaminant filtering rule altogether.

LeeBergstrand (changes requested):
Thanks. It looks like we are very close to finishing this branch. Good Job I highlighted a few more changes.

LeeBergstrand:
@jmtsuji I'm working on an elegant solution to the short-read contaminant set up.

rotary/rules/rotary.smk
LeeBergstrand: Addressed in https://github.com/jmtsuji/rotary/pull/82

rotary/rules/rotary.smk
LeeBergstrand: Addressed in https://github.com/jmtsuji/rotary/pull/82

rotary/rules/rotary.smk
LeeBergstrand: Addressed in https://github.com/jmtsuji/rotary/pull/82

rotary/config.yaml
LeeBergstrand: Implement in https://github.com/jmtsuji/rotary/pull/82

jmtsuji:
@LeeBergstrand Thanks for the proposed revision! I just added my comments at #82

jmtsuji:
@LeeBergstrand I think this PR is only waiting on #82, correct? If so, then feel free to merge this PR after finishing the #82 merge. Otherwise, feel free to keep this PR open if you'd like to do another code review.

PR #82, opened 2023-11-14
Short read improvements

LeeBergstrand:
- removed perform_contaminant_filtration config parameter. Users can now leave both lists empty to skip filtering.
- removed get_contamination_reference_file_paths as it is no longer necessary.
- add code for symlinking custom file paths to the database directory.

### DIFF
--- a/rotary/rules/rotary.smk
+++ b/rotary/rules/rotary.smk
@@ -523,8 +523,10 @@ rule short_read_contamination_filter:
     input:
         short_r1 = "{sample}/qc/short/{sample}_quality_trim_R1.fastq.gz",
         short_r2 = "{sample}/qc/short/{sample}_quality_trim_R2.fastq.gz",
-        ncbi_contamination_references = expand(os.path.join(DB_DIR_PATH, 'contamination_references', 'ncbi', '{accession}.fna.gz'), accession=CONTAMINATION_NCBI_ACCESSIONS),
-        custom_contamination_references = expand(os.path.join(DB_DIR_PATH, 'contamination_references', 'custom', '{contaminant_name}.fna.gz'), contaminant_name=CUSTOM_CONTAMINATION_FILE_NAMES)
+        ncbi_contamination_references = expand(os.path.join(DB_DIR_PATH, 'contamination_references', 'ncbi',
+            '{accession}.fna.gz'), accession=CONTAMINATION_NCBI_ACCESSIONS),
+        custom_contamination_references = expand(os.path.join(DB_DIR_PATH, 'contamination_references', 'custom',
+            '{contaminant_name}.fna.gz'), contaminant_name=CUSTOM_CONTAMINATION_FILE_NAMES)
     output:
         short_r1 = temp("{sample}/qc/short/{sample}_filter_R1.fastq.gz"),
         short_r2 = temp("{sample}/qc/short/{sample}_filter_R2.fastq.gz"),
@@ -538,6 +540,8 @@ rule short_read_contamination_filter:
         "{sample}/benchmarks/qc/short/short_read_contamination_filter.benchmark.txt"
     params:
         contamination_filter_kmer_length = config.get("contamination_filter_kmer_length"),
+        contamination_references = lambda wildcards, input: ','.join(input.ncbi_contamination_references +
+                                                                     input.custom_contamination_references)
     threads:
         config.get("threads", 1)
     resources:
@@ -545,7 +549,7 @@ rule short_read_contamination_filter:
     shell:
         """
         bbduk.sh -Xmx{resources.mem}g threads={threads} in={input.short_r1} in2={input.short_r2} \
-          ref={input.ncbi_contamination_references} {input.custom_contamination_references} out={output.short_r1} out2={output.short_r2} \
+          ref={params.contamination_references} out={output.short_r1} out2={output.short_r2} \
           stats={output.filter_stats} qhist={output.quality_histogram} \
           k={params.contamination_filter_kmer_length} ktrim=f rcomp=t \
           overwrite=t interleaved=f qin=33 pigz=t unpigz=t \
--- a/rotary/rules/rotary.smk
+++ b/rotary/rules/rotary.smk
@@ -8,7 +8,7 @@ import itertools
 from snakemake.utils import min_version
 
 from rotary.sample import parse_sample_tsv
-from rotary.utils import symlink_or_compress, get_contamination_reference_file_paths
+from rotary.utils import symlink_or_compress
 
 
 ZENODO_VERSION = "10087395"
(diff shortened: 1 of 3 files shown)


LeeBergstrand:
@jmtsuji This simplifies a bunch of your code. Thoughts? Note I have only '-dry-run' tested this.

rotary/rules/rotary.smk
```diff
@@ -518,16 +538,14 @@ rule short_read_contamination_filter:
...
-            config.get("contamination_references_custom_filepaths"))
     threads:
         config.get("threads", 1)
     resources:
         mem = config.get("memory")
     shell:
         """
         bbduk.sh -Xmx{resources.mem}g threads={threads} in={input.short_r1} in2={input.short_r2} \
-          ref={params.all_contaminant_references} out={output.short_r1} out2={output.short_r2} \
+          ref={input.ncbi_contamination_references} {input.custom_contamination_references} out={output.short_r1} out2={output.short_r2} \
```
LeeBergstrand, line 548: @jmtsuji Does bbduk need `,` between the file paths if there are multiple input files?

rotary/rules/rotary.smk
```diff
@@ -518,16 +538,14 @@ rule short_read_contamination_filter:
         "{sample}/benchmarks/qc/short/short_read_contamination_filter.benchmark.txt"
     params:
         contamination_filter_kmer_length = config.get("contamination_filter_kmer_length"),
-        all_contaminant_references = ','.join(CONTAMINANT_REFERENCE_GENOMES_NCBI +
```
LeeBergstrand, line 521: @jmtsuji If we need commas I think there is a ways to access input variables inside params. https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#non-file-parameters-for-rules
Maybe a lambda function?

rotary/rules/rotary.smk
```diff
@@ -19,13 +19,25 @@ START_HMM_NAME = os.path.splitext(os.path.basename(config.get("hmm_url")))[0]
...
 DB_DIR_PATH = config.get('db_dir')
-CONTAMINANT_REFERENCE_GENOMES_NCBI=get_contamination_reference_file_paths(
-    config.get("contamination_references_ncbi_accessions"), DB_DIR_PATH)
+
 SAMPLE_TSV_PATH = 'samples.tsv'
 SAMPLES = parse_sample_tsv(SAMPLE_TSV_PATH)
 
 SAMPLE_NAMES = list(SAMPLES.keys())
 
+CONTAMINATION_NCBI_ACCESSIONS = config.get("contamination_references_ncbi_accessions")
```
LeeBergstrand, line 28: @jmtsuji Note that once we start breaking the snakemake file into multi files we can move this logic to the top of the QC module.

rotary/rules/rotary.smk
jmtsuji: Indeed, commas are needed based on reading the docs

rotary/rules/rotary.smk
jmtsuji: Agreed -- this sounds like a good idea

rotary/rules/rotary.smk
```diff
@@ -98,6 +110,13 @@ rule download_contamination_reference:
         fi
         """
 
+rule set_up_custom_contamination_references:
+    output:
+        expand(os.path.join(DB_DIR_PATH, 'contamination_references', 'custom', '{contaminant_name}.fna.gz'),
+            contaminant_name=CUSTOM_CONTAMINATION_FILE_NAMES),
+    run:
+        for name, path in zip(CUSTOM_CONTAMINATION_FILE_NAMES,CUSTOM_CONTAMINATION_FILEPATHS):
+            symlink_or_compress(path, os.path.join(DB_DIR_PATH, 'contamination_references', 'custom', f'{name}.fna.gz'))
```
jmtsuji, line 119: This function assumes that the custom FastA files are gzipped. Do we want to require that the custom FastA files are gzipped? I think the BBDuk ref argument can handle FastA or gzipped FastA.

After checking if the input files are gzipped, we will either need to:
- throw an error if the files are not gzipped, 
- compress non-gzipped files, or
- handle their filepaths differently than gzipped files during symlinking and downstream steps

rotary/rules/rotary.smk
jmtsuji: Interesting -- this would be an elegant solution. Can you try this out?

rotary/rules/rotary.smk
```diff
@@ -19,13 +19,25 @@ START_HMM_NAME = os.path.splitext(os.path.basename(config.get("hmm_url")))[0]
...
 
+CONTAMINATION_NCBI_ACCESSIONS = config.get("contamination_references_ncbi_accessions")
+CUSTOM_CONTAMINATION_FILEPATHS = config.get("contamination_references_custom_filepaths")
+
+if CUSTOM_CONTAMINATION_FILEPATHS:
+    CUSTOM_CONTAMINATION_FILE_NAMES = [os.path.basename(path).split('.')[0] for path in CUSTOM_CONTAMINATION_FILEPATHS]
+else:
+    CUSTOM_CONTAMINATION_FILE_NAMES = []
+
+if not CONTAMINATION_NCBI_ACCESSIONS and not CUSTOM_CONTAMINATION_FILEPATHS:
```
jmtsuji, line 36: I agree, makes sense to move this logic here rather than have redundant variables in the user config file

rotary/rules/rotary.smk
```diff
@@ -19,13 +19,25 @@ START_HMM_NAME = os.path.splitext(os.path.basename(config.get("hmm_url")))[0]
...
 SAMPLE_TSV_PATH = 'samples.tsv'
 SAMPLES = parse_sample_tsv(SAMPLE_TSV_PATH)
 
 SAMPLE_NAMES = list(SAMPLES.keys())
 
+CONTAMINATION_NCBI_ACCESSIONS = config.get("contamination_references_ncbi_accessions")
+CUSTOM_CONTAMINATION_FILEPATHS = config.get("contamination_references_custom_filepaths")
+
+if CUSTOM_CONTAMINATION_FILEPATHS:
+    CUSTOM_CONTAMINATION_FILE_NAMES = [os.path.basename(path).split('.')[0] for path in CUSTOM_CONTAMINATION_FILEPATHS]
```
jmtsuji, line 32: As commented on below, we will need to consider how to handle gzipped files vs. non gzipped files.

jmtsuji (changes requested):
@LeeBergstrand Nice idea to simplify the contamination filter code. I left some comments, mostly about handling gzipped vs. non-gzipped custom FastA files and the need to have comma-separated input into the bbduk ref argument. Mind taking a look? Thanks!

rotary/rules/rotary.smk
LeeBergstrand: Fixed in https://github.com/jmtsuji/rotary/pull/82/commits/c6ff6557d46609d5fd28edf005c4bddaa22218f0

rotary/rules/rotary.smk
LeeBergstrand: @jmtsuji Isn't that what `symlink_or_compress` function does? If the file is already compressed, it symlinks it. It compresses and writes the compressed output to the desired path if it is not compressed. That was the function's intended purpose. Or am I missing something?

rotary/rules/rotary.smk
LeeBergstrand: Implemented.

rotary/rules/rotary.smk
LeeBergstrand: @jmtsuji, see this function and the function above it. Python can gzip files using the zlib library, the underlying code for gzip.

https://github.com/jmtsuji/rotary/blob/c6ff6557d46609d5fd28edf005c4bddaa22218f0/rotary/utils.py#L91

LeeBergstrand:
> @LeeBergstrand Nice idea to simplify the contamination filter code. I left some comments, mostly about handling gzipped vs. non-gzipped custom FastA files and the need to have comma-separated input into the bbduk ref argument. Mind taking a look? Thanks!

I thing this is already handled. See my comments to https://github.com/jmtsuji/rotary/pull/82#discussion_r1391874866

LeeBergstrand:
@jmtsuji It passes my tests and makes it past the poly polish stage. If you think the gzip issues are addressed we should be good to merge.

rotary/rules/rotary.smk
jmtsuji: @LeeBergstrand Sorry, this was totally an oversight on my end. I think I must have read this code too hastily and somehow missed `symlink_or_compress`. Yes, your previous function addresses the gzip issue, so the code is OK as-is. My apologies for the wild goose chase!

jmtsuji (commented):
Gzip comments are addressed -- thanks for these improvements, looks good to merge.

jmtsuji (approved):

PR #88, opened 2023-12-11
Set GTDB DB dir each time GTDB-Tk is run

### LINKED ISSUE #87: GTDB-Tk installation error
## Problem description
The first time rotary is run on a system, GTDB-Tk works fine. However, when rotary is re-run, if a re-install of the GTDB-Tk conda env is needed (e.g., due to a change in the [env YAML file](https://github.com/jmtsuji/rotary/blob/06e2be93bc63dca17cce12b0ac5834fd26a1e69f/rotary/envs/gtdbtk.yaml)), GTDB-Tk will fail to run due to the GTDB root dir not being set.

Version: Working with the current develop branch (around commit [06e2be9](https://github.com/jmtsuji/rotary/commit/06e2be93bc63dca17cce12b0ac5834fd26a1e69f))
System: Linux (Ubuntu 22.04)

## Possible cause
Rule [setup_gdtb](https://github.com/jmtsuji/rotary/blob/06e2be93bc63dca17cce12b0ac5834fd26a1e69f/rotary/rules/rotary.smk#L229C1-L229C17) sets the GTDB root dir (`GTDBTK_DATA_PATH`) in the GTDB-Tk conda env. After this rule is run, it saves a checkpoint file in the rotary DB dir (in `checkpoints/GTDB_[VERSION_GTDB_COMPLETE]_setup`). Because the setup file is saved in the rotary DB dir and not in the individual run dir, it means that the setup rule will never be run again if that DB dir is used, even if a new GTDB-Tk conda env is installed. Thus, new GTDB-Tk conda envs will not have the GTDB root dir set.

## Possible solutions
This problem should be pretty easy to address. Here are two possible solutions:
- Make rule setup_gtdb save the checkpoint file in the rotary run (output) dir. This way, the rule will be re-run each time that rotary is run.
- Delete the setup_gtdb rule and instead set `GTDBTK_DATA_PATH` in [rule run_gtdbtk](https://github.com/jmtsuji/rotary/blob/06e2be93bc63dca17cce12b0ac5834fd26a1e69f/rotary/rules/rotary.smk#L1397). This way, the GTDB DB path will be set just before GTDB-Tk is run.

The only case where I could see one of these working better than the other is if a different program was trying to use the same GTDB-Tk conda env with a different GTDB dir. In that case, it would be better to set `GTDBTK_DATA_PATH` just before GTDB-Tk is run (solution 2 above) to avoid accidentally using the wrong GTDB. However, I imagine this kind of conflict would be very rare. It would be ideal if `GTDBTK_DATA_PATH` could be set as a flag of GTDB-Tk, but I don't see an option to do this.

@LeeBergstrand What's your favoured approach? Thanks!

jmtsuji:
Fixes #87 via proposed solution 2

### DIFF
--- a/rotary/rules/rotary.smk
+++ b/rotary/rules/rotary.smk
@@ -226,32 +226,10 @@ rule download_gtdb_db:
         touch {output.install_finished}
         """
 
-rule setup_gtdb:
-    input:
-        os.path.join(DB_DIR_PATH,"checkpoints","GTDB_" + VERSION_GTDB_COMPLETE + "_download")
-    output:
-        os.path.join(DB_DIR_PATH,"checkpoints","GTDB_" + VERSION_GTDB_COMPLETE + "_setup")
-    conda:
-        "../envs/gtdbtk.yaml"
-    log:
-        "logs/download/gtdb_db_setup.log"
-    benchmark:
-        "benchmarks/download/gtdb_db_setup.txt"
-    params:
-        db_dir=os.path.join(DB_DIR_PATH,"GTDB_" + VERSION_GTDB_COMPLETE),
-    shell:
-        """
-        # Set the database path in the conda installation
-        echo "Set: GTDBTK_DATA_PATH={params.db_dir}" > {log}
-        conda env config vars set GTDBTK_DATA_PATH={params.db_dir}
-
-        touch {output}
-        """
-
 
 rule validate_gtdb:
     input:
-        os.path.join(DB_DIR_PATH,"checkpoints","GTDB_" + VERSION_GTDB_COMPLETE + "_setup")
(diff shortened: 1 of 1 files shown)


jmtsuji:
@LeeBergstrand After thinking about it, I lean toward the 2nd solution I proposed in issue #87. In case it's helpful, I went ahead and made a PR with this proposed solution. Feel free to provide your comments or to close this PR if you feel a different approach would be better. Thanks!

jmtsuji:
This fix passes for `rule validate_gtdb`. I haven't yet tested it for `rule run_gtdbtk`, but the updated syntax should work the same way for both rules (a test is running now).

LeeBergstrand (approved):

LeeBergstrand:
@jmtsuji I did a little editing of the spacing in run gtdbtk. But overall, it looks good. merging.

jmtsuji:
Great, thanks!

Issue #92, opened 2023-12-12
Specify more temp files


jmtsuji:
Rotary generates a lot of files that seem extraneous. We should add more of these to the outputs of rotary rules and make them temp().

LeeBergstrand:
![temp_files](https://github.com/rotary-genomics/rotary/assets/5819462/2ed7bb2f-eeeb-4247-9a71-9f20d3b7aeab)

@jmtsuji I used snakemake's file graph command to make this image. Temped files are highlighted in red, and my suggestions are highlighted in green. Thoughts?

LeeBergstrand:
I was also thing we could temp all the polishing steps. There's a few rules here where we kept output diaganotic txt output files. I think we could add a rule to compile these to TSVs and then we would be able to temp out their entire directories.

LeeBergstrand:
Here are the folder sizes of the output:

25M     ./assembly/end_repair                                                                                                                                                                                                    
14M     ./assembly/flye/20-repeat                                                                                                                                                                                                
5.5M    ./assembly/flye/00-assembly                                                                                                                                                                                              
17M     ./assembly/flye/30-contigger                                                                                                                                                                                             
13M     ./assembly/flye/40-polishing                                                                                                                                                                                             
5.7M    ./assembly/flye/10-consensus                                                                                                                                                                                             
66M     ./assembly/flye                                                                                                                                                                                                          
90M     ./assembly                                                                                                                                                                                                               
189M    ./qc/long                                                                                                                                                                                                                
143M    ./qc/short                                                                                                                                                                                                               
331M    ./qc                                                                                                                                                                                                                     
9.5M    ./polish/polypolish/input                                                                                                                                                                                                
239M    ./polish/polypolish                                                                                                                                                                                                      
16M     ./polish/polca                                                                                                                                                                                                           
422M    ./polish/medaka                                                                                                                                                                                                          
16M     ./polish/cov_filter
691M    ./polish
2.3M    ./annotation/checkm/protein_files
2.7M    ./annotation/checkm
3.1M    ./annotation/eggnog
310M    ./annotation/coverage
6.7M    ./annotation/dfast/ddbj
6.9M    ./annotation/dfast/genbank
73M     ./annotation/dfast
388M    ./annotation
5.6M    ./circularize/combine
9.5M    ./circularize/polypolish/input
239M    ./circularize/polypolish
12M     ./circularize/circlator
5.6M    ./circularize/filter
261M    ./circularize

[...]

PR #100, opened 2023-12-15
delete requirement to check for other polca files.

LeeBergstrand:
I ran into this error where Polca wouldn't find the files:

- {sample}_polypolish.fasta.unSorted.sam"
- {sample}_polypolish.fasta.alignSorted.bam"

The polca folder has the other files created by the polca run but not these. Does the newer version of Polca not produce these files? Either way, none of these files are used by the follow-on steps, so I removed them from the required outputs.

### DIFF
--- a/rotary/rules/rotary.smk
+++ b/rotary/rules/rotary.smk
@@ -800,22 +800,21 @@ rule polish_polca:
         qc_short_r2 = "{sample}/qc/{sample}_qc_R2.fastq.gz",
         polished = "{sample}/polish/polypolish/{sample}_polypolish.fasta"
     output:
-        polca_output = "{sample}/polish/polca/{sample}_polca.fasta"
+        polca_output = temp("{sample}/polish/polca/{sample}_polca.fasta"),
+        outdir = temp(directory("{sample}/polish/polca"))
     conda:
         "../envs/masurca.yaml"
     log:
         "{sample}/logs/polish/polca.log"
     benchmark:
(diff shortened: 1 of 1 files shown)


jmtsuji:
@LeeBergstrand Interesting that those files were not generated for you. I wonder if we could add a code block like the following to `shell` in that rule, so that those files are deleted, if they are made?

(`polypolish_sam` and `polypolish_bam` would be moved to params)
```bash
if [[ -f {params.polypolish_sam} ]]; then
  rm {params.polypolish_sam}
fi

if [[ -f {params.polypolish_bam} ]]; then
  rm {params.polypolish_bam}
fi
```
This code could be added underneath the final `cd` in the `shell` block. Thoughts?

jmtsuji:
From @LeeBergstrand in #102 :

> POLCA leaves an entire folder full of output files not specified in the rule output. I'm considering having it delete its entire folder using temp. See: https://github.com/rotary-genomics/rotary/issues/92

Agreed that just specifying the whole output directory as a temp file might work really well, given that the output files from this rule might be a bit unstable. Shall we pursue this option?

LeeBergstrand:
@jmtsuji Added code to delete the entire polca directory between runs: https://github.com/rotary-genomics/rotary/commit/e8e237555cc9cfbb44f9906c6822534e685dc105

There was an issue where if polca were rerun on a prior run's data with an existing polca folder, then it would cause errors like the above.

jmtsuji:
@LeeBergstrand This makes sense -- thanks!

Issue #90, opened 2023-12-12
Split rotary snakefile into multiple files


jmtsuji:
To make rotary's code more understandable, we should split the current `rotary.smk` into multiple snakefiles, for example:

- main: `rotary.smk`
- download: `download.smk` (with DB versions)
- long read qc: `qc_long.smk`
- short read qc: `qc_short.smk`
- assembly: `assembly.smk`
- polishing: `polish.smk`
- circularization: `circularize.smk`
- annotation: `annotate.smk`

jmtsuji:
@LeeBergstrand Are you okay to do this after we merge the current develop into master?

LeeBergstrand:
@jmtsuji Sure!

jmtsuji:
Addressed in #114 (for some reason, linking the PR to the issue doesn't seem to be closing the issue automatically once the PR is merged)

PR #104, opened 2023-12-22
split snakemake file into multiple.

LeeBergstrand:
Addresses https://github.com/rotary-genomics/rotary/issues/90

Split rotary into multiple snakemake files.


LeeBergstrand:
@jmtsuji Split the snakemake file into multiple. Can you double-check that all the rules are in the right places?

PR #110, opened 2024-01-02
Refactor snakefile

LeeBergstrand:
- split snakemake file into multiple files.
- add more temp files or use Shadow not to save excess files


LeeBergstrand:
@jmtsuji Can you review to determine if the Snakemakefile was broken up properly? In other words the rules are in the correct files?

LeeBergstrand:
I decided to merge all the snakemake refactoring into one branch so both breaking it up into multiple files and also doing the temps and cleaning up the number of output files etc.

Issue #59, opened 2023-10-20
Allow for mixture of samples with or without short-read data.


LeeBergstrand:
I see -- sounds tricky to modify the logic at the moment. It might work well to revisit having an alternative way to parse the file names (as a low priority feature request) at the same time as changing the code to allow for optional short read data, at some point in the future. Sounds good to set this as a long-term TODO.

_Originally posted by @jmtsuji in https://github.com/jmtsuji/rotary/issues/53#issuecomment-1771928862_

jmtsuji:
@LeeBergstrand I wonder if some of the new semantic helper functions for inputs in snakemake 8 might help with the per-sample switching logic? See https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#the-lookup-function and https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#the-branch-function  For example, using the `lookup` function, we could check if the params for a specific sample show that it is a long-read only sample. Hybrid assembly-specific rules could still be run for the long-read only samples but could have a built-in "bypass" conditional that causes the inputs to be pushed through without changes, if the params show that a sample is long-read only. Alternatively, `branch` could be used to create files for hybrid samples in a special sub-folder that is then subject to hybrid assembly rules. Thoughts? (I am just jotting down some brief notes quickly.)

I hope soon to get back to working on the internal code to replace circlator... this will allow us to upgrade to snakemake 8. We might also be able to temporarily disable circlator on a test branch to try out snakemake 8 before I am finished with the circlator replacement code. Let me know if you think the above semantic helper functions might be of use -- I haven't run any tests yet.

jmtsuji:
@LeeBergstrand I think we decided that, for now, running a mix of samples with hybrid vs. long-read-only datasets is a low priority use case. Users can just run rotary in separate batches of samples with hybrid vs. long read data if needed. Feel free to re-open this issue if you feel it needs to be addressed more urgently.

Issue #45, opened 2023-10-17
Create Rotary Download CLI Command


LeeBergstrand:
Yeah, that would make sense to add a `rotary download` command eventually. I think this should be pretty straightforward:
- All commands that involve DB downloads are summarized at the top of the snakefile at the moment.
- An additional checkpoint-like rule (e.g., `rule download`) could be added to summarize the outputs from these rules. 
- Then, it would be easy to call this `download` rule via snakemake using the `--until` flag, with very similar code as would be used in `rotary run`.

_Originally posted by @jmtsuji in https://github.com/jmtsuji/rotary/issues/39#issuecomment-1765922554_

jmtsuji:
@LeeBergstrand Check this out: https://snakemake.readthedocs.io/en/stable/snakefiles/deployment.html

> Conda deployment also works well for offline or air-gapped environments. Running snakemake --use-conda --conda-create-envs-only will only install the required conda environments without running the full workflow. Subsequent runs with --use-conda will make use of the local environments without requiring internet access.

We could potentially add `snakemake --use-conda --conda-create-envs-only` as part of the download CLI so that all conda envs are also set up using this command. Thoughts?

LeeBergstrand:
@jmtsuji This sounds great. We could add this to the downloads sub-command. However, we should check to see if we need to have the input pipeline files set up beforehand.

jmtsuji:
@LeeBergstrand Good point -- if the input pipeline files are needed, then this flag might not be as useful for our case. I suppose we could try to force it to work by providing end-to-end test files or something. Let's see what the behaviour is like when we have a chance to test it.

LeeBergstrand:
Further Discussed in https://github.com/rotary-genomics/rotary/pull/114#issuecomment-1936528046

[...]

PR #114, opened 2024-01-27
Develop lee

### LINKED ISSUE #90: Split rotary snakefile into multiple files

### LINKED ISSUE #92: Specify more temp files

LeeBergstrand:
Refactor many aspects of the rotary code base to support further extension and development.

Primary Changes:

- Split the rotary.smk pipeline into multiple files that are specialized for each group of tasks done by the pipeline. For example, all the QC rules are placed in a qc.smk file. The rotary.smk file now includes these other files when running the pipeline.
- Refactor and rewrite the code for handling the creation of sample TSV files to make them more extensible. Different types of sequencing files or file pairs (e.g., long or paired short reads) are represented as objects. These objects can be grouped into sample objects representing the multiple sequencing files used to represent one physical sample. Several sample objects represent different types of samples (e.g., long read sample, paired short read sample, or long with polishing reads samples). A dataset object represents multiple samples contained in a single sequencing run. Dataset objects containing multiple sample objects are used to either generate sample TSVs or can be generated by sample TSVs.
- I refactored the CLI code for Rotary to use the above objects and moved the code for running SnakeMake as a subprocess into its own file.
- Addressed: https://github.com/rotary-genomics/rotary/issues/116


LeeBergstrand:
@jmtsuji Feel free to review when you get a chance. The eventual intent would be to move the following files to their software library and GitHub repository within the Rotary organization. The code could be used to build other snake make pipelines.

- run.py
- sample.py
- dataset.py
- utils.py

My intent would be to use this library for our own internal snake-make pipelines at my work.

jmtsuji:
> @jmtsuji Feel free to review when you get a chance. The eventual intent would be to move the following files to their software library and GitHub repository within the Rotary organization. The code could be used to build other snake make pipelines.
> 
> * run.py
> * sample.py
> * dataset.py
> * utils.py
> 
> My intent would be to use this library for our own internal snake-make pipelines at my work.

Reviewing now -- thanks for your patience! Yes, that sounds good to plan to move the files you mentioned to their own software library / repo.

rotary/dataset.py
```diff
@@ -0,0 +1,220 @@
...
+    for file in fastq_files:
+        identifier = file.identifier
+        if identifier in samples_files.keys():
+            samples_files[identifier].append(file)
+        else:
+            samples_files[identifier] = [file]
+
+    samples = []
+    for identifier, sequencing_files in samples_files.items():
+        if len(sequencing_files) != 3:
```
jmtsuji, line 188: Given that `auto_create_sample_from_files` can support long-read only configurations, am I correct that, in the long term, the `generate_dataset_from_fastq_directory` function can be extended fairly easily to include long-read only samples?

rotary/rules/polish.smk
```diff
@@ -0,0 +1,360 @@
...
+    output:
+        "{sample}/polish/cov_filter/{sample}_filtered_contigs.list"
+    params:
+        meandepth_short=config.get("meandepth_cutoff_short_read"),
+        evenness_short=config.get("evenness_cutoff_short_read"),
+        meandepth_long=config.get("meandepth_cutoff_long_read"),
+        evenness_long=config.get("evenness_cutoff_long_read")
+    run:
+        #
+        def filter_coverage_data(coverage_file, mean_depth, evenness):
```
jmtsuji, line 270: Thanks for improving the function definition here. How do you think we should handle such functions? Might it be better to move functions like this into a different part of the snakefile (or to a different file in the repo) rather than keep it nested inside the rule?

rotary/rules/annotation.smk
```diff
@@ -0,0 +1,399 @@
...
+        stats="{sample}/stats"
+    run:
+        source_relpath = os.path.relpath(str(params.logs),os.path.dirname(str(output.logs)))
+        os.symlink(source_relpath,str(output.logs))
+
+        source_relpath = os.path.relpath(str(params.stats),os.path.dirname(str(output.stats)))
+        os.symlink(source_relpath,str(output.stats))
+
+
+# TODO - can I remove the use of cd?
```
jmtsuji: Nice that you could simplify this code and eliminate the need for `cd` (with AI)! I think we can remove these comments above the `summarize_annotation` rule now.

rotary/rules/annotation.smk
```diff
@@ -0,0 +1,399 @@
...
+
+if POLISH_WITH_SHORT_READS == True:
+    # TODO - clarify name compared to previous mapping step
+    rule calculate_final_short_read_coverage:
+        input:
+            qc_short_r1="{sample}/qc/{sample}_qc_R1.fastq.gz",
+            qc_short_r2="{sample}/qc/{sample}_qc_R2.fastq.gz",
+            dfast_genome="{sample}/annotation/dfast/{sample}_genome.fna"
+        output:
+            mapping=temp("{sample}/annotation/coverage/{sample}_short_read.bam"),
```
jmtsuji, line 297: I personally like keeping the BAM files for the final long and short read coverage calculations, because users can then check how reads mapped in specific places to look for abnormalities. However, from a high-throughput perspective, the BAM files waste a lot of space. What would you think about adding a setting in the config (e.g., `keep_bam_files`) that users can turn on so that the final BAM files are not treated as `temp()`? We could leave the setting False by default. This config setting could be used in a rule like this:
```python
mapping="{sample}/annotation/coverage/{sample}_short_read.bam" if keep_bam_files else temp("{sample}/annotation/coverage/{sample}_short_read.bam")
``` 

This would apply to `calculate_final_short_read_coverage` and `calculate_final_long_read_coverage`.

rotary/rules/assembly.smk
```diff
@@ -0,0 +1,112 @@
...
+# SAMPLE_NAMES is instantiated in rotary.smk
+
+rule assembly_flye:
+    input:
+        "{sample}/qc/{sample}_qc_long.fastq.gz"
+    output:
+        assembly="{sample}/assembly/flye/{sample}_assembly.fasta",
+        info="{sample}/assembly/flye/{sample}_assembly_info.txt",
+        output_dir=directory("{sample}/assembly/flye"),
+        assembly_dir=temp(directory("{sample}/assembly/flye/00-assembly")),
```
jmtsuji, line 15: Thanks for adding these directories as temp()

rotary/rules/polish.smk
```diff
@@ -0,0 +1,360 @@
...
+    input:
+        "{sample}/polish/medaka/{sample}_consensus.fasta"
+    output:
+        temp("{sample}/polish/polypolish/input/{sample}_input.fasta")
+    run:
+        source_relpath = os.path.relpath(str(input),os.path.dirname(str(output)))
+        os.symlink(source_relpath,str(output))
+
+
+read_mapping_file_extensions = ['amb', 'ann', 'bwt', 'pac', 'sa']
```
jmtsuji: This is a nice idea. I wonder if this global var should be moved to the top of the snakefile?

rotary/rules/annotation.smk
```diff
@@ -0,0 +1,399 @@
...
+        """
+
+# TODO - I might need to remove special characters from strain name to use for locus tag prefix
+# TODO - can I auto-predict genome completeness, names, types, topologies?
+rule run_dfast:
+    input:
+        contigs="{sample}/circularize/{sample}_circularize.fasta",
+        install_finished=os.path.join(DB_DIR_PATH,"checkpoints","dfast_" + VERSION_DFAST)
+    output:
+        dfast_genome="{sample}/annotation/dfast/{sample}_genome.fna",
```
jmtsuji, line 155: Thanks for coding in these output file names!

jmtsuji (changes requested):
@LeeBergstrand The improvements look great! Thanks for your hard work making the general library files for setting up a snakemake run, splitting the snakefile into smaller chunks, adding more temp files, and making various other minor enhancements/fixes. I checked the split snakefiles, and all the rules seem to be in the right places. I have added a few minor comments of things to revise, but otherwise I think this PR is OK to merge.

One bigger picture question, though, is how we should handle download-related rules. Right now, they are added into the same snakefile as the analysis rules that need them. Another option would be to make a separate download.smk file and to put all the download rules in there. We could potentially set up a new rule `rule download` that can be optionally called by the CLI (but otherwise goes uncalled) in that snakefile. This would allow users to download all needed DB files in a single command if they wanted to, as per #45 . If you like this idea to make a download-dedicated snakefile, we could go ahead and implement it now, while we are already re-organizing the snakefiles, or we could wait until some point in the future... either is OK for me. What do you think?

Thanks again!

P.S. Note that I have not run an end-to-end test yet on this branch... I can do so if you would like.

LeeBergstrand:
> @LeeBergstrand The improvements look great! Thanks for your hard work making the general library files for setting up a snakemake run, splitting the snakefile into smaller chunks, adding more temp files, and making various other minor enhancements/fixes. I checked the split snakefiles, and all the rules seem to be in the right places. I have added a few minor comments of things to revise, but otherwise I think this PR is OK to merge.
> 
> One bigger picture question, though, is how we should handle download-related rules. Right now, they are added into the same snakefile as the analysis rules that need them. Another option would be to make a separate download.smk file and to put all the download rules in there. We could potentially set up a new rule `rule download` that can be optionally called by the CLI (but otherwise goes uncalled) in that snakefile. This would allow users to download all needed DB files in a single command if they wanted to, as per #45 . If you like this idea to make a download-dedicated snakefile, we could go ahead and implement it now, while we are already re-organizing the snakefiles, or we could wait until some point in the future... either is OK for me. What do you think?
> 
> Thanks again!
> 
> P.S. Note that I have not run an end-to-end test yet on this branch... I can do so if you would like.

From a maintenance standpoint, it makes more sense to me to have the database download rules in the same snakemake file as the rule that uses the data. That way, if you need to update anything, such as database versions, everything is self-contained in the same place so that you can edit both rules simultaneously. For example, when you specify a new model version name in the download rule and that same model version name in the processing rule, you can add both as a variable at the top of the file. Also, if you need to port things between software programs, everything is in one package. You can copy the file over. I would still add a `download.smk` file with the rule `rule download` but it would be the only thing in that file. This rule would be pretty long because it would check for the output files of all the download rules in the other files. Perhaps you could also have sub-rules (e.g., `qc_download`) in `download.smk` that check individual download steps and checkpoints). Because of how Snakemake works, you don't need to have all the download rules in `download.smk`. You could either import `download.smk` into `rotary.smk` and call the rule out of `rotary.smk` or you could import all the other rules into 'download.smk' and call `download.smk`

LeeBergstrand:
> > @LeeBergstrand The improvements look great! Thanks for your hard work making the general library files for setting up a snakemake run, splitting the snakefile into smaller chunks, adding more temp files, and making various other minor enhancements/fixes. I checked the split snakefiles, and all the rules seem to be in the right places. I have added a few minor comments of things to revise, but otherwise I think this PR is OK to merge.
> > One bigger picture question, though, is how we should handle download-related rules. Right now, they are added into the same snakefile as the analysis rules that need them. Another option would be to make a separate download.smk file and to put all the download rules in there. We could potentially set up a new rule `rule download` that can be optionally called by the CLI (but otherwise goes uncalled) in that snakefile. This would allow users to download all needed DB files in a single command if they wanted to, as per #45 . If you like this idea to make a download-dedicated snakefile, we could go ahead and implement it now, while we are already re-organizing the snakefiles, or we could wait until some point in the future... either is OK for me. What do you think?
> > Thanks again!
> > P.S. Note that I have not run an end-to-end test yet on this branch... I can do so if you would like.
> 
> From a maintenance standpoint, it makes more sense to me to have the database download rules in the same snakemake file as the rule that uses the data. That way, if you need to update anything, such as database versions, everything is self-contained in the same place so that you can edit both rules simultaneously. For example, when you specify a new model version name in the download rule and that same model version name in the processing rule, you can add both as a variable at the top of the file. Also, if you need to port things between software programs, everything is in one package. You can copy the file over. I would still add a `download.smk` file with the rule `rule download` but it would be the only thing in that file. This rule would be pretty long because it would check for the output files of all the download rules in the other files. Perhaps you could also have sub-rules (e.g., `qc_download`) in `download.smk` that check individual download steps and checkpoints). Because of how Snakemake works, you don't need to have all the download rules in `download.smk`. You could either import `download.smk` into `rotary.smk` and call the rule out of `rotary.smk` or you could import all the other rules into 'download.smk' and call `download.smk`

```Python
# Include various modules.
include: './qc.smk'
include: './assembly.smk'
include: './polish.smk'
include: './circularize.smk'
include: './annotation.smk'
```

Snakemake uses the old-school way of importing that C or PHP uses. This command copy-pastes the file's contents of each child snakemake file into the parent snakemake file at the point of the input statement before it is run. That's why I can call variables in the parent file that aren't declared in the child file. Ultimately, it just looks like one big file for the DAG code.

rotary/rules/annotation.smk
LeeBergstrand: @jmtsuji Sounds like a good idea. Will do.

LeeBergstrand:
> P.S. Note that I have not run an end-to-end test yet on this branch... I can do so if you would like.

I've been running end to end tests on this code for a while now. I think you can run it after we move it into dev.

rotary/dataset.py
LeeBergstrand: @jmtsuji Yes, the plan is to eventually have it so that it will be extended to support long read-only configurations. Once the sample object code is moved to its library, it will be extended a bit to support a variety of sample types. For example, my library at our company will be short read-only for 16S sequencing.

LeeBergstrand: > am I correct that, in the long term, the generate_dataset_from_fastq_directory function can be extended fairly easily to include long-read only samples?

This will be the eventual plan, but it will be tricky to handle, especially if it's a mix of samples with and without polishing reads. If our code can handle mixed long and hybrid types, we must rewrite the snake make code to process the samples one at a time or dynamically generate the DAG pathing. I can do long read-only right now, but I have to make two fake Illumina files and turn the short read polish flag in the config to `False`.

LeeBergstrand: > I can do long read-only right now, but I have to make two fake Illumina files and turn the short read polish flag in the config to `False.`

We can modify the rule `rule set_up_sample_directories:` with some config-dependent switching logic and not need the short reads. This would be the low-hanging fruit.

rotary/rules/polish.smk
LeeBergstrand: @jmtsuji, do you have to move imports inside the run block as if it's a script? I suspect everything outside the run block can't be seen inside unless it is a particular variable created by Snakemake. That is my suspicion, but I am unsure; it could be injecting it differently than expected. If we wanted to move it, I would move it to `utils.py` (most of utils.py would be moved to its library, but we can have two util files) or a dedicated `coverage` so as not to clutter up to code.

LeeBergstrand:
@jmtsuji, can we move your remaining suggestions to issues and merge the existing code we have right now? Because the code has changed so much, it's in our best interest to merge things ASAP so we can work off the same codebase and not drift too far apart. I can address your other issues on individual branches of primary development.

rotary/dataset.py
LeeBergstrand: Moved to https://github.com/rotary-genomics/rotary/issues/118

rotary/rules/annotation.smk
LeeBergstrand: Moved to https://github.com/rotary-genomics/rotary/issues/119

LeeBergstrand:
> > @LeeBergstrand The improvements look great! Thanks for your hard work making the general library files for setting up a snakemake run, splitting the snakefile into smaller chunks, adding more temp files, and making various other minor enhancements/fixes. I checked the split snakefiles, and all the rules seem to be in the right places. I have added a few minor comments of things to revise, but otherwise I think this PR is OK to merge.
> > One bigger picture question, though, is how we should handle download-related rules. Right now, they are added into the same snakefile as the analysis rules that need them. Another option would be to make a separate download.smk file and to put all the download rules in there. We could potentially set up a new rule `rule download` that can be optionally called by the CLI (but otherwise goes uncalled) in that snakefile. This would allow users to download all needed DB files in a single command if they wanted to, as per #45 . If you like this idea to make a download-dedicated snakefile, we could go ahead and implement it now, while we are already re-organizing the snakefiles, or we could wait until some point in the future... either is OK for me. What do you think?
> > Thanks again!
> > P.S. Note that I have not run an end-to-end test yet on this branch... I can do so if you would like.
> 
> From a maintenance standpoint, it makes more sense to me to have the database download rules in the same snakemake file as the rule that uses the data. That way, if you need to update anything, such as database versions, everything is self-contained in the same place so that you can edit both rules simultaneously. For example, when you specify a new model version name in the download rule and that same model version name in the processing rule, you can add both as a variable at the top of the file. Also, if you need to port things between software programs, everything is in one package. You can copy the file over. I would still add a `download.smk` file with the rule `rule download` but it would be the only thing in that file. This rule would be pretty long because it would check for the output files of all the download rules in the other files. Perhaps you could also have sub-rules (e.g., `qc_download`) in `download.smk` that check individual download steps and checkpoints). Because of how Snakemake works, you don't need to have all the download rules in `download.smk`. You could either import `download.smk` into `rotary.smk` and call the rule out of `rotary.smk` or you could import all the other rules into 'download.smk' and call `download.smk`

Moved my comments to: https://github.com/rotary-genomics/rotary/issues/45

rotary/rules/polish.smk
jmtsuji: Moved this discussion to #120 - I think this can wait until we merged in this PR.

rotary/dataset.py
jmtsuji: Cool that this can be extended to a variety of different configurations in the future! Agreed, for rotary, we're going to have to modify the snakefile before we can easily accommodate long read only inputs (see #59 ). For low-hanging fruit, we can 1. make sure that rotary can finish successfully when the short read polish flag is `False` and 2. work on the `set_up_sample_directories` rule like you mentioned. I'll also post another idea in #59 about how to handle the switching logic. I'll resolve this conversation for now... we can continue to discuss in #118 and #59

rotary/rules/annotation.smk
jmtsuji: Sounds good to continue in #119! (I fixed what I think was a typo in your last post... I think you were discussing 119 rather than 118) Resolving this conversation for now.

jmtsuji:
> I've been running end to end tests on this code for a while now. I think you can run it after we move it into dev.

OK, sounds good!

> @jmtsuji, can we move your remaining suggestions to issues and merge the existing code we have right now? Because the code has changed so much, it's in our best interest to merge things ASAP so we can work off the same codebase and not drift too far apart. I can address your other issues on individual branches of primary development.

Yes, that sounds fine to move my remaining suggestions to issues and merge this PR ASAP. I've moved the remaining discussions in this PR to issues. It looks like everything else is resolved. Feel free to merge once ready!

PR #115, opened 2024-02-05
Refactor sample

LeeBergstrand:
Refactor sample code to use a new, expandable series of data objects to represent datasets and various types of samples.


LeeBergstrand (commented):
Looks good.

PR #122, opened 2024-02-13
Install Polypolish via Conda

LeeBergstrand:
I noticed that we were doing a manual install of poly-polish while working on creating a unified download rule. Decided to simplify the code.

- Create a new conda environment to install polypolish.
  - Move to polypolish 0.6.0
- Remove the rule for installing polypolish.
- Simplify rule `polish_polypolish`:
  - Remove inputs to polypolish scripts.
  - Call polypolish sub commands directly.

### DIFF
--- a/rotary/rules/polish.smk
+++ b/rotary/rules/polish.smk
@@ -6,33 +6,12 @@ import sys
 
 import pandas as pd
 
-VERSION_POLYPOLISH="0.5.0"
 
 DB_DIR_PATH = config.get('db_dir')
 
 READ_MAPPING_FILE_EXTENSIONS = ['amb', 'ann', 'bwt', 'pac', 'sa']
 
-
 # SAMPLE_NAMES and POLISH_WITH_SHORT_READS are instantiated in rotary.smk
-
-rule install_polypolish:
-    output:
-        polypolish_filter=os.path.join(DB_DIR_PATH,"polypolish_" + VERSION_POLYPOLISH,"polypolish_insert_filter.py"),
-        polypolish=os.path.join(DB_DIR_PATH,"polypolish_" + VERSION_POLYPOLISH,"polypolish"),
-        install_finished=os.path.join(DB_DIR_PATH,"checkpoints","polypolish_" + VERSION_POLYPOLISH)
-    log:
-        "logs/download/install_polypolish.log"
-    benchmark:
-        "benchmarks/download/install_polypolish.txt"
-    params:
-        db_dir=os.path.join(DB_DIR_PATH,"polypolish_" + VERSION_POLYPOLISH),
-        url="https://github.com/rrwick/Polypolish/releases/download/v" + VERSION_POLYPOLISH + "/polypolish-linux-x86_64-musl-v" + VERSION_POLYPOLISH + ".tar.gz"
-    shell:
-        """
-        mkdir -p {params.db_dir}
-        wget -O - {params.url} 2> {log} | tar -C {params.db_dir} -xzf - >> {log} 2>&1
-        touch {output.install_finished}
-        """
 
 rule prepare_medaka_polish_input:
     input:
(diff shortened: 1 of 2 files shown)


LeeBergstrand:
@jmtsuji I'm running tests on it now. Was there a specific reason why you installed poly-polish outside the conda before, or was there not a conda package ready when writing the original code?

rotary/rules/polish.smk
```diff
@@ -118,11 +92,11 @@ rule polish_polypolish:
...
         printf "\n\n### Polypolish insert filter ###\n" >> {log}
-        {input.polypolish_filter} --in1 {output.mapping_r1} --in2 {output.mapping_r2} \
+        polypolish filter --in1 {output.mapping_r1} --in2 {output.mapping_r2} \
           --out1 {output.mapping_clean_r1} --out2 {output.mapping_clean_r2} 2>> {log}
 
         printf "\n\n### Polypolish ###\n" >> {log}
-        {input.polypolish} {input.contigs} --debug {output.debug} \
+        polypolish polish --debug {output.debug} {input.contigs}  \
           {output.mapping_clean_r1} {output.mapping_clean_r2} 2>> {log} | 
           seqtk seq -A -l 0 |
```
jmtsuji: Polypolish 0.6.0 changed now it modifies sequence header names compared to polypolish 0.5.0: https://github.com/rrwick/Polypolish/releases/tag/v0.6.0 (It used to add `_polypolish` to the ends of all sequences, but now it adds ` polypolish`.)

I think we can eliminate the `awk` code line entirely now. Also, we can add the `-C` flag to `seats seq` to drop the header comments. Sounds good?

jmtsuji (changes requested):
Nice find! There wasn't a conda install for polypolish before. Yes, let's replace the manual polypolish install with the conda install like you've done here. I added one comment about handling how polypolish 0.6.0 changes sequence headers, but otherwise, if this branch passes tests, feel free to merge this PR. Thanks!

jmtsuji:
One more thought: if we wanted, we could separate the read mapping step for polypolish from polypolish itself. The read mapping step can be highly parallelized, whereas I think polypolish only runs with a single thread. This doesn't have to be done now, but it could be a nice optimization.

LeeBergstrand:
> One more thought: if we wanted, we could separate the read mapping step for polypolish from polypolish itself. The read mapping step can be highly parallelized, whereas I think polypolish only runs with a single thread. This doesn't have to be done now, but it could be a nice optimization.

Sounds good. I decided to move this to its own issue. I'm going to be doing some modifications to read mapping and I'll roll these changes into it.

jmtsuji:
> > One more thought: if we wanted, we could separate the read mapping step for polypolish from polypolish itself. The read mapping step can be highly parallelized, whereas I think polypolish only runs with a single thread. This doesn't have to be done now, but it could be a nice optimization.
> 
> Sounds good. I decided to move this to its own issue. I'm going to be doing some modifications to read mapping and I'll roll these changes into it.

Sounds good!

LeeBergstrand:
> > > One more thought: if we wanted, we could separate the read mapping step for polypolish from polypolish itself. The read mapping step can be highly parallelized, whereas I think polypolish only runs with a single thread. This doesn't have to be done now, but it could be a nice optimization.
> > 
> > 
> > Sounds good. I decided to move this to its own issue. I'm going to be doing some modifications to read mapping and I'll roll these changes into it.
> 
> Sounds good!

Will fix when I add this improvement: https://github.com/rotary-genomics/rotary/issues/124

LeeBergstrand:
@jmtsuji Requesting re-review. I combined the two `seqtk seq` commands rather than piping them into each other. Should that work? The code appears to be passing my tests.

jmtsuji (approved):
Thanks for catching the repeated use of `seqtk`! That was a leftover from fixing the sequence header names in previous polypolish versions. It should be no problem to simplify to a single line of seqtk. Feel free to merge this PR once ready.

PR #125, opened 2024-02-14
Improve read mapping rules

### LINKED ISSUE #119: Add flag to keep read coverage mapping BAM files.
I personally like keeping the BAM files for the final long and short read coverage calculations, because users can then check how reads mapped in specific places to look for abnormalities. However, from a high-throughput perspective, the BAM files waste a lot of space. What would you think about adding a setting in the config (e.g., `keep_bam_files`) that users can turn on so that the final BAM files are not treated as `temp()`? We could leave the setting False by default. This config setting could be used in a rule like this:
```python
mapping="{sample}/annotation/coverage/{sample}_short_read.bam" if keep_bam_files else temp("{sample}/annotation/coverage/{sample}_short_read.bam")
``` 

This would apply to `calculate_final_short_read_coverage` and `calculate_final_long_read_coverage`.

_Originally posted by @jmtsuji in https://github.com/rotary-genomics/rotary/pull/114#discussion_r1482665387_

### LINKED ISSUE #123: Breakout polypolish read mapping to its own rule.
One more thought: if we wanted, we could separate the read mapping step for polypolish from polypolish itself. The read mapping step can be highly parallelized, whereas I think polypolish only runs with a single thread. This doesn't have to be done now, but it could be a nice optimization.

_Originally posted by @jmtsuji in https://github.com/rotary-genomics/rotary/issues/122#issuecomment-1940404380_

### LINKED ISSUE #124: Simplify read mapping file extension code in rules.
Problem Description:

There is a more straightforward to account for multiple extensions rather than the code below. 

```python
read_mapping_files= temp(expand("{{sample}}/{{step}}/polypolish/input/{{sample}}_input.fasta.{ext}", ext=READ_MAPPING_FILE_EXTENSIONS))
```

Problem Solution:

Use this instead:

```python
read_mapping_files= temp(multiext("{sample}/{step}/polypolish/input/{{sample}}_input.fasta", *READ_MAPPING_FILE_EXTENSIONS))
```

LeeBergstrand:
- pull read mapping code out of rule `polish_polypolish` and move it to a new rule.
- utilize the `multiext()` function for removing read mapping index files (improved readability and caching).
- move from `bwa` to `bwa-mem2`, the latest `bwa` mapping algorithm (faster and much smaller index on disk)
- trick `polca.sh` into using `bwa-mem2` by symlinking `bwa-mem2` to `bwa`.
- add config flag to keep read mapping bam files used final read coverage statistics.

### DIFF
--- a/rotary/rules/polish.smk
+++ b/rotary/rules/polish.smk
@@ -150,10 +150,18 @@ rule polish_polca:
         mem=config.get("memory")
     shell:
         """
+        printf "\n\n### Replace BWA with bwa-mem2 ###\n" >> {log}
+        bwa_mem2_path=$(which bwa-mem2)
+        bwa_mem2_dir=$(dirname $bwa_mem2_path)
+        ln -s $bwa_mem2_path $bwa_mem2_dir/bwa
+        
+        printf "\n\n### Run POLCA ###\n" >> {log}
         cd {params.outdir}
         polca.sh -a ../../../{input.polished} -r "../../../{input.qc_short_r1} ../../../{input.qc_short_r2}" -t {threads} -m {resources.mem}G > ../../../{log} 2>&1
         ln -s "{wildcards.sample}_polypolish.fasta.PolcaCorrected.fa" "{wildcards.sample}_polca.fasta"
         cd ../../../
+        
+        printf "\n\n### Done. ###\n"
         """
 
 
--- a/rotary/rules/polish.smk
+++ b/rotary/rules/polish.smk
@@ -155,7 +155,7 @@ rule polish_polca:
         if [ -z "${{bwa_path}}" ]; then
             bwa_mem2_path="$(which bwa-mem2)"
             bwa_mem2_dir="$(dirname "${{bwa_mem2_path}}")"
-            ln -s "${{bwa_mem2_path}}" "${{bwa_mem2_dir/bwa}}"
+            ln -s "${{bwa_mem2_path}}" "${{bwa_mem2_dir}}/bwa"
         
             for variant in avx avx2 avx512bw sse41 sse42; do
                 bwa_mem2_variant_path="$(which bwa-mem2.${{variant}})"
--- a/rotary/rules/polish.smk
+++ b/rotary/rules/polish.smk
@@ -151,14 +151,17 @@ rule polish_polca:
     shell:
         """
         printf "\n\n### Replace bwa with bwa-mem2 ###\n" >> {log}
-        bwa_mem2_path=$(which bwa-mem2)
-        bwa_mem2_dir=$(dirname $bwa_mem2_path)
-        ln -s $bwa_mem2_path $bwa_mem2_dir/bwa
+        bwa_path=$(which bwa)
+        if [ -z "$bwa_path" ]; then
+            bwa_mem2_path=$(which bwa-mem2)
+            bwa_mem2_dir=$(dirname $bwa_mem2_path)
+            ln -s $bwa_mem2_path $bwa_mem2_dir/bwa
         
-        for variant in avx avx2 avx512bw sse41 sse42; do
-            bwa_mem2_variant_path=$(which bwa-mem2.$variant)
-            ln -s bwa_mem2_variant_path $bwa_mem2_dir/bwa.$variant
-        done
+            for variant in avx avx2 avx512bw sse41 sse42; do
+                bwa_mem2_variant_path=$(which bwa-mem2.$variant)
+                ln -s $bwa_mem2_variant_path $bwa_mem2_dir/bwa.$variant
+            done
+        fi
         
         printf "\n\n### Run POLCA ###\n" >> {log}
         cd {params.outdir}
--- a/rotary/rules/polish.smk
+++ b/rotary/rules/polish.smk
@@ -150,16 +150,16 @@ rule polish_polca:
         mem=config.get("memory")
     shell:
         """
-        printf "\n\n### Replace bwa with bwa-mem2 ###\n" >> {log}
-        bwa_path=$(which bwa)
-        if [ -z "$bwa_path" ]; then
-            bwa_mem2_path=$(which bwa-mem2)
-            bwa_mem2_dir=$(dirname $bwa_mem2_path)
-            ln -s $bwa_mem2_path $bwa_mem2_dir/bwa
+        printf "### Replace bwa with bwa-mem2 ###\n" >> {log}
+        bwa_path="$(which bwa)"
+        if [ -z "${{bwa_path}}" ]; then
+            bwa_mem2_path="$(which bwa-mem2)"
+            bwa_mem2_dir="$(dirname "${{bwa_mem2_path}}")"
+            ln -s "${{bwa_mem2_path}}" "${{bwa_mem2_dir/bwa}}"
         
             for variant in avx avx2 avx512bw sse41 sse42; do
-                bwa_mem2_variant_path=$(which bwa-mem2.$variant)
-                ln -s $bwa_mem2_variant_path $bwa_mem2_dir/bwa.$variant
+                bwa_mem2_variant_path="$(which bwa-mem2.${{variant}})"
+                ln -s "${{bwa_mem2_variant_path}}" "${{bwa_mem2_dir}}/bwa.${{variant}}"
             done
         fi
         
--- a/rotary/rules/polish.smk
+++ b/rotary/rules/polish.smk
@@ -9,7 +9,7 @@ import pandas as pd
 
 DB_DIR_PATH = config.get('db_dir')
 
-READ_MAPPING_FILE_EXTENSIONS = ['amb', 'ann', 'bwt', 'pac', 'sa']
+READ_MAPPING_FILE_EXTENSIONS = ['.amb', '.ann', '.bwt', '.pac', '.sa']
 
 # SAMPLE_NAMES and POLISH_WITH_SHORT_READS are instantiated in rotary.smk
 
@@ -30,7 +30,7 @@ rule polish_medaka:
     output:
         dir=directory("{sample}/{step}/medaka"),
         contigs="{sample}/{step}/medaka/{sample}_consensus.fasta",
-        calls_to_draft=temp(expand("{{sample}}/{{step}}/medaka/calls_to_draft.{ext}", ext=['bam', 'bam.bai'])),
+        calls_to_draft=temp(multiext("{sample}/{step}/medaka/calls_to_draft", '.bam', '.bam.bai')),
         consensus_probs=temp("{sample}/{step}/medaka/consensus_probs.hdf")
     conda:
         "../envs/medaka.yaml"
@@ -68,8 +68,8 @@ rule map_short_reads_for_polishing:
     output:
         mapping_r1 = temp("{sample}/{step}/polypolish/{sample}_R1.sam"),
         mapping_r2 = temp("{sample}/{step}/polypolish/{sample}_R2.sam"),
-        read_mapping_files= temp(expand("{{sample}}/{{step}}/polypolish/input/{{sample}}_input.fasta.{ext}",
-            ext=READ_MAPPING_FILE_EXTENSIONS))
+        read_mapping_files= temp(multiext("{sample}/{step}/polypolish/input/{sample}_input.fasta",
+            *READ_MAPPING_FILE_EXTENSIONS))
     conda:
         "../envs/mapping.yaml"
     log:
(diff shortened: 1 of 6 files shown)


LeeBergstrand:
@jmtsuji I just made some significant changes to the read mapping rules. I am looking forward to your review.

rotary/rules/polish.smk
```diff
@@ -135,10 +150,26 @@ rule polish_polca:
         mem=config.get("memory")
     shell:
         """
+        printf "\n\n### Replace bwa with bwa-mem2 ###\n" >> {log}
```
jmtsuji: I just noticed that a standalone version of POLCA, re-implemented in Python, was released that is called `pypolca`: https://github.com/gbouras13/pypolca (a [conda install](https://bioconda.github.io/recipes/pypolca/README.html#package-pypolca) also seems to exist)
This is now the recommended tool for polishing after Polypolish, in place of POLCA: https://github.com/rrwick/Polypolish/wiki
It also lets you set the output directory, unlike the original POLCA.
Looks like this was made by the hybracter team as part of their hybrid assembly pipeline.

What would you think about using this in place of POLCA?
It still relies on bwa, but the code you've written here could be used to swap out bwa for bwa-mem2 (or we might be able to directly raise an issue with the tool's authors to add bwa-mem2 support).

rotary/rules/polish.smk
```diff
@@ -135,10 +150,26 @@ rule polish_polca:
...
+        bwa_path=$(which bwa)
```
jmtsuji: I know that only some parts of our current snakemake code are robust to whitespaces, but I think it might make sense to make this bwa -> bwa-mem2 code whitespace robust. `$(which bwa)` would become `"$(which bwa)"`, and so on. If you'd like, I'd be happy to push a commit that implements this -- let me know!

rotary/rules/polish.smk
jmtsuji: One additional thought: because POLCA/pypolca was built with bwa (not bwa-mem2), I wonder if its polishing results could change a little after swapping to bwa-mem2. Do you think it might be worthwhile to do some benchmarking before swapping to bwa-mem2 specifically for POLCA/pypolca? I think the other changes to bwa-mem2 you've made in this PR all look like they should have limited to no impact to pipeline results.

jmtsuji (changes requested):
Thanks for these additions! Everything looks good, except I've flagged a couple things in the POLCA section, including a new idea and a question. Look forward to your feedback.

rotary/rules/polish.smk
LeeBergstrand: > I just noticed that a standalone version of POLCA, re-implemented in Python, was released that is called `pypolca`: https://github.com/gbouras13/pypolca (a [conda install](https://bioconda.github.io/recipes/pypolca/README.html#package-pypolca) also seems to exist) This is now the recommended tool for polishing after Polypolish, in place of POLCA: https://github.com/rrwick/Polypolish/wiki It also lets you set the output directory, unlike the original POLCA. Looks like this was made by the hybracter team as part of their hybrid assembly pipeline.

@jmtsuji Let's move over to `pypolca`. Looking at the original `polca`'s github repo it looks like it was developed using poor software development practices. I was going to ask you about a replacement for stability reasons. So, let's go with the reimplementation.

LeeBergstrand: > What would you think about using this in place of POLCA? It still relies on bwa, but the code you've written here could be used to swap out bwa for bwa-mem2 (or we might be able to directly raise an issue with the tool's authors to add bwa-mem2 support).

@jmtsuji I would keep my code in for now and then raise the issue with the devs simutaneously.

rotary/rules/polish.smk
LeeBergstrand: > I know that only some parts of our current snakemake code are robust to whitespaces, but I think it might make sense to make this bwa -> bwa-mem2 code whitespace robust. `$(which bwa)` would become `"$(which bwa)"`, and so on. If you'd like, I'd be happy to push a commit that implements this -- let me know!

@jmtsuji Feel free to fix this!

rotary/rules/polish.smk
LeeBergstrand: > One additional thought: because POLCA/pypolca was built with bwa (not bwa-mem2), I wonder if its polishing results could change a little after swapping to bwa-mem2. Do you think it might be worthwhile to do some benchmarking before swapping to bwa-mem2 specifically for POLCA/pypolca? I think the other changes to bwa-mem2 you've made in this PR all look like they should have limited to no impact to pipeline results.

@jmtsuji You would get changes between vanilla bwa code versions as well. For example, last year they inserted a bug into BTW that slightly changed the read mapping results in a way that was not intended. Also this would affect poly-polish as well since it takes similar read maps for input.

> > I just noticed that a standalone version of POLCA, re-implemented in Python, was released that is called `pypolca`: https://github.com/gbouras13/pypolca (a [conda install](https://bioconda.github.io/recipes/pypolca/README.html#package-pypolca) also seems to exist) This is now the recommended tool for polishing after Polypolish, in place of POLCA: https://github.com/rrwick/Polypolish/wiki It also lets you set the output directory, unlike the original POLCA. Looks like this was made by the hybracter team as part of their hybrid assembly pipeline.
> 
> @jmtsuji Let's move over to `pypolca`. Looking at the original `polca`'s github repo it looks like it was developed using poor software development practices. I was going to ask you about a replacement for stability reasons. So, let's go with the reimplementation.

@jmtsuji Note that the two tools produce slightly different results: https://github.com/gbouras13/pypolca/blob/main/benchmarking.md

LeeBergstrand: > > One additional thought: because POLCA/pypolca was built with bwa (not bwa-mem2), I wonder if its polishing results could change a little after swapping to bwa-mem2. Do you think it might be worthwhile to do some benchmarking before swapping to bwa-mem2 specifically for POLCA/pypolca? I think the other changes to bwa-mem2 you've made in this PR all look like they should have limited to no impact to pipeline results.
> 
> @jmtsuji You would get changes between vanilla bwa code versions as well. For example, last year they inserted a bug into BTW that slightly changed the read mapping results in a way that was not intended. Also this would affect poly-polish as well since it takes similar read maps for input.
> 
> > > I just noticed that a standalone version of POLCA, re-implemented in Python, was released that is called `pypolca`: https://github.com/gbouras13/pypolca (a [conda install](https://bioconda.github.io/recipes/pypolca/README.html#package-pypolca) also seems to exist) This is now the recommended tool for polishing after Polypolish, in place of POLCA: https://github.com/rrwick/Polypolish/wiki It also lets you set the output directory, unlike the original POLCA. Looks like this was made by the hybracter team as part of their hybrid assembly pipeline.
> > 
> > 
> > @jmtsuji Let's move over to `pypolca`. Looking at the original `polca`'s github repo it looks like it was developed using poor software development practices. I was going to ask you about a replacement for stability reasons. So, let's go with the reimplementation.
> 
> @jmtsuji Note that the two tools produce slightly different results: https://github.com/gbouras13/pypolca/blob/main/benchmarking.md

I think the polka workflow goes: BWA --> Samtools --> freebayes --> some filtering and post analysis. The software isn't fundamentally reliant on BWA. Theoretically, if it was programmed differently, it could take SAM/BAM files as input rather than taking the reads directly taking the raw reads. If it were implemented that way, it would be similar to polypolish and we could probably use the same read mapping step between the two of them (I would suggest this to the `pyploca` dev).

LeeBergstrand: > I think the polka workflow goes: BWA --> Samtools --> freebayes --> some filtering and post analysis. The software isn't fundamentally reliant on BWA. Theoretically, if it was programmed differently, it could take SAM/BAM files as input rather than taking the reads directly taking the raw reads. If it were implemented that way, it would be similar to polypolish and we could probably use the same read mapping step between the two of them (I would suggest this to the `pyploca` dev).
>
> Do you think it might be worthwhile to do some benchmarking before swapping to bwa-mem2 specifically for POLCA/pypolca? 

It might be worth some benchmarking or asking the `pyploca` dev.

rotary/rules/polish.smk
LeeBergstrand: @jmtsuji Feel free to make any changes necessary to update this branch.

rotary/rules/polish.smk
jmtsuji: @LeeBergstrand OK, let's switch over to `pypolca`! (I also was also thinking it would be nice to eventually switch away from using POLCA.) I will leave it up to you whether we should make the switch in this PR versus wait until a future PR.

That also makes sense to ask the `pypolca` devs about adding bwa-mem2 support or allowing read mapping to be separated from the rest of `pypolca`. The `pypolca` devs might be able to re-run their existing benchmarks you linked me to, which would save us time of trying to make similar benchmarks ourselves. Good point that even changing bwa versions (or other tool versions) can affect the results. I'll make an issue in the `pypolca` repo.

Also, I agree, let's keep the code you wrote for swapping bwa and bwa-mem2 in this PR. We can use this for the time being.

rotary/rules/polish.smk
jmtsuji: Just made some updates and pushed them into this PR. I haven't tested the updated rule but ran some basic bash syntax checks in command line to confirm that the general types of edits I made were OK. Hope this helps!

rotary/rules/polish.smk
LeeBergstrand: > @LeeBergstrand OK, let's switch over to `pypolca`! (I also was also thinking it would be nice to eventually switch away from using POLCA.) I will leave it up to you whether we should make the switch in this PR versus wait until a future PR.
> 
> That also makes sense to ask the `pypolca` devs about adding bwa-mem2 support or allowing read mapping to be separated from the rest of `pypolca`. The `pypolca` devs might be able to re-run their existing benchmarks you linked me to, which would save us time of trying to make similar benchmarks ourselves. Good point that even changing bwa versions (or other tool versions) can affect the results. I'll make an issue in the `pypolca` repo.
> 
> Also, I agree, let's keep the code you wrote for swapping bwa and bwa-mem2 in this PR. We can use this for the time being.

@jmtsuji I sent you some additional comments in a Signal DM.

LeeBergstrand:
>OK, let's switch over to pypolca! (I also was also thinking it would be nice to eventually switch away from using POLCA.) I will leave it up to you whether we should make the switch in this PR versus wait until a future PR.

@jmtsuji Can you do the swap out to `pypolca` in this PR?

jmtsuji:
> > OK, let's switch over to pypolca! (I also was also thinking it would be nice to eventually switch away from using POLCA.) I will leave it up to you whether we should make the switch in this PR versus wait until a future PR.
> 
> @jmtsuji Can you do the swap out to `pypolca` in this PR?

As per our DM discussion, let's keep the bwa-mem2 replacement code and continue to use POLCA for the time being. We can look into swapping to pypolca in future -- I'll create a separate issue for it. I imagine that bwa-mem2 will probably make little to no impact on the POLCA results, but I'll at least benchmark this on one example genome for now, when I am able to do so... I'll hold off on merging this PR until I have a change to run this single genome test. As we discussed, I'll take over the final merge of this PR.

LeeBergstrand:
> > > OK, let's switch over to pypolca! (I also was also thinking it would be nice to eventually switch away from using POLCA.) I will leave it up to you whether we should make the switch in this PR versus wait until a future PR.
> > 
> > 
> > @jmtsuji Can you do the swap out to `pypolca` in this PR?
> 
> As per our DM discussion, let's keep the bwa-mem2 replacement code and continue to use POLCA for the time being. We can look into swapping to pypolca in future -- I'll create a separate issue for it. I imagine that bwa-mem2 will probably make little to no impact on the POLCA results, but I'll at least benchmark this on one example genome for now, when I am able to do so... I'll hold off on merging this PR until I have a change to run this single genome test. As we discussed, I'll take over the final merge of this PR.

Sounds good!

LeeBergstrand:
@jmtsuji Are we ready to merge this yet or are you still working on it?

jmtsuji:
@LeeBergstrand Still working on it -- I'd like to do a quick single-genome benchmark of BWA vs BWA-MEM2 with POLCA before merging, if that's okay.

jmtsuji:
@LeeBergstrand I compared POLCA with BWA vs. BWA-MEM2 using the "_Ca._ Chx. allophototropha" genome. POLCA was run after polishing of the genome by Polypolish. The resulting polished genome files were identical whether BWA or BWA-MEM2 was used. Because of this, I will go ahead and merge this PR.

Issue #118, opened 2024-02-09
Allow long read-only with when polish_with_short_reads set to 'False'.


LeeBergstrand:
> I can do long read-only right now, but I have to make two fake Illumina files and turn the short read polish flag in the config to `False.`

We can modify the rule `rule set_up_sample_directories:` with some config-dependent switching logic and not need the short reads. This would be the low-hanging fruit.

_Originally posted by @LeeBergstrand in https://github.com/rotary-genomics/rotary/pull/114#discussion_r1484858455_

LeeBergstrand:
@jmtsuji rule `summarize_contigs_by_coverage` in `polish.smk` fails when there is no short reads.

LeeBergstrand:
It refers to both long and short rules in the same rule.

LeeBergstrand:
@jmtsuji Fixed by https://github.com/rotary-genomics/rotary/pull/146

Issue #120, opened 2024-02-13
Simplify Python code within certain snakemake rules


jmtsuji:
Some snakemake rules (e.g., rule `rule summarize_contigs_by_coverage` in polish.smk) currently include long Python code blocks that contain functions etc. Would it be better to simplify the code included in the Snakefile, e.g., by calling a separate Python script, for such code?

jmtsuji:
Previous discussion for `rule summarize_contigs_by_coverage` in #114:

@jmtsuji 
> Thanks for improving the function definition here. How do you think we should handle such functions? Might it be better to move functions like this into a different part of the snakefile (or to a different file in the repo) rather than keep it nested inside the rule?

@LeeBergstrand 
> @jmtsuji, do you have to move imports inside the run block as if it's a script? I suspect everything outside the run block can't be seen inside unless it is a particular variable created by Snakemake. That is my suspicion, but I am unsure; it could be injecting it differently than expected. If we wanted to move it, I would move it to utils.py (most of utils.py would be moved to its library, but we can have two util files) or a dedicated coverage so as not to clutter up to code.

jmtsuji:
@LeeBergstrand Regarding your comment above, I think you are probably right... I also suspect that everything outside the code block can't be seen inside. We will probably need to import the code via an `import` command or just write a Python script that we can call for the rule.

jmtsuji:
Given that we don't have very many rules with long Python code blocks at the moment, I think it's not a high priority to try to move around the code at this point. However, it might be something we could consider in the future, especially if we create more rules like `summarize_contigs_by_coverage` that have complex Python code. OK to table this issue for now? Or do you think it could be worthwhile to explore simplifying the code in the snakefile for this rule?

LeeBergstrand:
> Given that we don't have very many rules with long Python code blocks at the moment, I think it's not a high priority to try to move around the code at this point. However, it might be something we could consider in the future, especially if we create more rules like `summarize_contigs_by_coverage` that have complex Python code. OK to table this issue for now? Or do you think it could be worthwhile to explore simplifying the code in the snakefile for this rule?

I would table the issue for now. I would probably move the code to `utils.py` but after we pull the the sample object code out into its own library.

jmtsuji:
OK, sounds good -- let's table this for now.

jmtsuji:
[...]

PR #126, opened 2024-02-14
Fix coverage summary

jmtsuji:
Fixes a couple intertwined issues in rule `summarize_contigs_by_coverage` that caused errors when running rotary in long read only mode, to address one part of #118 

In the previous version of rotary, long-read only mode actually worked with rule `summarize_contigs_by_coverage`, but only when a coverage filter threshold was supplied in the config file. The rule resulted in an error when no long read coverage filter thresholds were supplied (which is the config default) in long-read only mode.

In this update, the following changes were made to address this problem:
- rule `summarize_contigs_by_coverage` was moved inside a conditional block (as it should have always been) so that it only runs when some kind of coverage filtration is requested
- a basic report summarizing the results of coverage filtration was created -- this is symlinked to `{sample}/polish/{sample}_contig_info.tsv`. This report is made whether coverage filtration is performed or not.
- the above report is used to parse out which contigs to run circularization on in checkpoint `split_circular_and_linear_contigs` -- before, no report was made if coverage filtration was not run, causing this checkpoint to fail.

Because of how rule `summarize_contigs_by_coverage` was moved into an existing conditional block in `polish.smk`, the git diff does not show the file changes very cleanly -- my apologies!

In the long term, I think it might make sense to move the code in `summarize_contigs_by_coverage` into a Python utility script -- this would make it easier to write/run tests on some of the logic being performed in this code.

Let me know if you have any thoughts -- thanks!

### DIFF
--- a/rotary/rules/assembly.smk
+++ b/rotary/rules/assembly.smk
@@ -16,8 +16,7 @@ rule assembly_flye:
         consensus_dir=temp(directory("{sample}/assembly/flye/10-consensus")),
         repeat_dir=temp(directory("{sample}/assembly/flye/20-repeat")),
         contigger_dir=temp(directory("{sample}/assembly/flye/30-contigger")),
-        polishing_dir=temp(directory("{sample}/assembly/flye/40-polishing")),
-
+        polishing_dir=temp(directory("{sample}/assembly/flye/40-polishing")) if config.get("flye_polishing_rounds") > 0 else []
     conda:
         "../envs/assembly_flye.yaml"
     log:
@@ -34,7 +33,7 @@ rule assembly_flye:
     shell:
         """
         flye --{params.input_mode} {input} {params.read_error} --out-dir {output.output_dir} {params.meta_mode} \
-          --iterations {params.polishing_rounds} -t {threads} > {log} 2>&1 
+          --iterations {params.polishing_rounds} -t {threads} > {log} 2>&1
         mv {output.output_dir}/assembly.fasta {output.assembly} 
         mv {output.output_dir}/assembly_info.txt {output.info} 
         """
--- a/rotary/rules/circularize.smk
+++ b/rotary/rules/circularize.smk
@@ -37,8 +37,8 @@ checkpoint split_circular_and_linear_contigs:
     output:
         directory("{sample}/circularize/filter/lists")
     run:
-        contig_info = pd.read_csv(input,sep='\t')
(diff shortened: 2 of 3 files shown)


rotary/rules/assembly.smk
```diff
@@ -16,8 +16,7 @@ rule assembly_flye:
         consensus_dir=temp(directory("{sample}/assembly/flye/10-consensus")),
         repeat_dir=temp(directory("{sample}/assembly/flye/20-repeat")),
         contigger_dir=temp(directory("{sample}/assembly/flye/30-contigger")),
-        polishing_dir=temp(directory("{sample}/assembly/flye/40-polishing")),
-
+        polishing_dir=temp(directory("{sample}/assembly/flye/40-polishing")) if config.get("flye_polishing_rounds") > 0 else []
```
jmtsuji, line 19: I made this change to fix a separate issue I ran into during testing -- if polishing is turned off for Flye (e.g., to save speed during testing), then Flye does not produce the output folder `40-polishing`

rotary/rules/circularize.smk
```diff
@@ -33,18 +33,15 @@ rule download_hmm:
 # Based on clustering tutorial at https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html (accessed 2022.3.31)
 checkpoint split_circular_and_linear_contigs:
     input:
-        assembly_stats="{sample}/assembly/{sample}_circular_info.tsv",
-        filter_list="{sample}/polish/cov_filter/{sample}_filtered_contigs.list"
+        "{sample}/polish/{sample}_contig_info.tsv"
```
jmtsuji, line 36: This checkpoint now uses the new contig summary report made during coverage filtration.

LeeBergstrand: Great

LeeBergstrand:
> In the long term, I think it might make sense to move the code in `summarize_contigs_by_coverage` into a Python utility script -- this would make it easier to write/run tests on some of the logic being performed in this code.

@jmtsuji This sounds good. Feel free to create an issue for this.

jmtsuji:
> > In the long term, I think it might make sense to move the code in `summarize_contigs_by_coverage` into a Python utility script -- this would make it easier to write/run tests on some of the logic being performed in this code.
> 
> @jmtsuji This sounds good. Feel free to create an issue for this.

Will include in issue #120

LeeBergstrand (approved):

LeeBergstrand:
@jmtsuji Looks good to me at a glance. Feel free to merge in.

Issue #132, opened 2024-02-15
QC read files deleted too early


jmtsuji:
# Run / system info
I'm running rotary commit 40381cc on Ubuntu 22.04.3, within the snakemake argument `--until circularize` (to run everything up until the start of the annotation module).

# Problem description
After reaching `checkpoint split_circular_and_linear_contigs`, I encountered the following error:

```
BUG: Out of jobs ready to be started, but not all files built yet. Please check https://github.com/snakemake/snakemake/issues/823 for more information.
Remaining jobs:
 - search_contig_start: S18/circularize/identify/S18_circular.faa, S18/circularize/identify/S18_circular.ffn, S18/circularize/identify/S18_circular.gff, S18/circularize/identify/S18_hmmsearch_hits.txt, S18/circularize/identify/S18_hmmsearch_hits_no_comments.txt
 - finalize_circular_contig_rotation: S18/circularize/combine/S18_circular.fasta
 - combine_circular_and_linear_contigs: S18/circularize/combine/S18_combined.fasta
 - process_start_genes: S18/circularize/identify/S18_start_genes.list
 - polish_polypolish: S18/circularize/polypolish/S18_R1.clean.sam, S18/circularize/polypolish/S18_R2.clean.sam, S18/circularize/polypolish/S18_polypolish.fasta, S18/circularize/polypolish/polypolish.debug.log, S18/stats/circularize/polypolish_changes.log
 - map_short_reads_for_polishing: S18/circularize/polypolish/S18_R1.sam, S18/circularize/polypolish/S18_R2.sam, S18/circularize/polypolish/input/S18_input.fasta.0123, S18/circularize/polypolish/input/S18_input.fasta.amb, S18/circularize/polypolish/input/S18_input.fasta.ann, S18/circularize/polypolish/input/S18_input.fasta.bwt.2bit.64, S18/circularize/polypolish/input/S18_input.fasta.pac
 - circularize: checkpoints/circularize
 - get_start_genes: S18/circularize/identify/S18_start_gene.ffn
 - prepare_polypolish_circularize_input: S18/circularize/polypolish/input/S18_input.fasta
 - run_circlator: S18/circularize/circlator/S18_rotated.fasta, S18/circularize/circlator, S18/circularize/circlator/rotated.promer.contigs_with_ends.fa
 - symlink_circularization: S18/circularize/S18_circularize.fasta
 - get_polished_contigs: S18/circularize/filter/S18_circular.fasta
[...]

PR #135, opened 2024-02-19
Replace POLCA with pypolca

jmtsuji:
As discussed in #125 , pypolca was recently developed as a standalone replacement for POLCA. This PR implements pypolca in place of POLCA in rotary for increased stability and multi-OS support.


rotary/config.yaml
```diff
@@ -93,8 +93,12 @@ keep_unrepaired_contigs: 'True'
 medaka_model: "r941_min_sup_g507"
 # Batch size: related to memory usage; can reduce if you get OOM (out of memory) errors
 medaka_batch_size: 100
-# Should the pipeline polish with short reads using polypolish.
+# Should the pipeline polish with short reads using Polypolish and pypolca?
 polish_with_short_reads: 'True'
+# Set Polypolish and pypolca to --careful mode, for short read datasets with mean coverage of about 20-25x or less?
+#  This will make polishing more convervative; Polypolish will not necessarily correct errors in repeat regions, and
+#  pypolca will use a more conservative threshold for determining when to change a DNA base pair
+careful_short_read_polishing: 'False'
```
jmtsuji, line 101: Note to self: test changing this setting to True and make sure everything still runs OK.

rotary/rules/circularize.smk
```diff
@@ -245,7 +245,7 @@ rule prepare_polypolish_circularize_input:
     input:
         "{sample}/circularize/circlator/{sample}_rotated.fasta"
     output:
-        "{sample}/circularize/polypolish/input/{sample}_input.fasta"
+        temp("{sample}/circularize/polypolish/input/{sample}_input.fasta")
```
jmtsuji, line 248: Why is this Polypolish input rule temp() and the other one (`prepare_polypolish_polish_input`) not temp()? I remember tinkering with these while trying to ensure that the pipeline finished without bugs related to the DAG, but I am not sure I compared them carefully. Because these are symlinks, the safe thing to do would be to keep both as non-temp, unless I have a chance to test in more detail.

rotary/rules/annotation.smk
```diff
@@ -309,16 +309,18 @@ if POLISH_WITH_SHORT_READS == True:
             "{sample}/logs/annotation/calculate_final_short_read_coverage.log"
         benchmark:
             "{sample}/benchmarks/annotation/calculate_final_short_read_coverage.txt"
+        params:
+            mem_per_thread = int(config.get("memory") / config.get("threads",1))
```
jmtsuji, line 313: Note that I generally changed how rules with samtools handle memory. Before, the mem per thread value (required by samtools) was set in `resources`. However, I realized that this could cause a bug with resource management, because to snakemake, this implies that the max. memory that rule will use is the GB per thread, which is incorrect. Now, the global mem is specified in `resources`, and then the mem per thread is calculated in `params`.

rotary/rules/polish.smk
```diff
@@ -122,61 +120,53 @@ rule polish_polypolish:
...
         
-        printf "\n\n### Run POLCA ###\n" >> {log}
-        cd {params.outdir}
-        polca.sh -a ../../../{input.polished} -r "../../../{input.qc_short_r1} ../../../{input.qc_short_r2}" -t {threads} -m {resources.mem}G > ../../../{log} 2>&1
-        ln -s "{wildcards.sample}_polypolish.fasta.PolcaCorrected.fa" "{wildcards.sample}_polca.fasta"
-        cd ../../../
+        mv {params.report} {output.report}
         
-        printf "\n\n### Done. ###\n"
+        # Delete the logfile generated by pypolca that has a timestamp on the end - we already save another log file
```
jmtsuji, line 155: I had to add this somewhat long code block to identify and remove a copy of the log file, saved by pypolca, that I wanted to be temp. The log file is saved with the date/time stamp on the end, so it seemed like snakemake was unable to identify it as an `output` (due to the wildcard timestamp).

rotary/rules/circularize.smk
jmtsuji: Removed the temp() on this output file now.

rotary/config.yaml
jmtsuji: Tested -- finishes without error, and repair stats look OK

jmtsuji:
@LeeBergstrand Mind giving your review on this PR to change POLCA to pypolca? Thanks!

rotary/rules/polish.smk
```diff
@@ -55,7 +53,7 @@ rule prepare_polypolish_polish_input:
     input:
         "{sample}/polish/medaka/{sample}_consensus.fasta"
     output:
-        temp("{sample}/polish/polypolish/input/{sample}_input.fasta")
+        "{sample}/polish/polypolish/input/{sample}_input.fasta"
```
LeeBergstrand: @jmtsuji Was removing `temp()` on purpose here?

rotary/rules/polish.smk
```diff
@@ -122,61 +120,53 @@ rule polish_polypolish:
...
+rule polish_pypolca:
     input:
         qc_short_r1 = "{sample}/qc/{sample}_qc_R1.fastq.gz",
         qc_short_r2 = "{sample}/qc/{sample}_qc_R2.fastq.gz",
         polished = "{sample}/polish/polypolish/{sample}_polypolish.fasta"
     output:
-        polca_output = temp("{sample}/polish/polca/{sample}_polca.fasta"),
-        polypolish_sam = temp("{sample}/polish/polca/{sample}_polypolish.fasta.unSorted.sam"),
-        polypolish_bam = temp("{sample}/polish/polca/{sample}_polypolish.fasta.alignSorted.bam"),
-        read_mapping_files= temp(multiext("{sample}/polish/polca/{sample}_polypolish.fasta.bwa",
```
LeeBergstrand, line 135: @jmtsuji Does pyploca clean up its BWA files after it runs? If not we should leave some of these removed parameters in place so Snakemake deletes the BWA output files.

rotary/rules/polish.smk
LeeBergstrand: Great!

LeeBergstrand (changes requested):

LeeBergstrand:
@jmtsuji Looks excellent so far. Just a few minor questions to address, and we should be able to merge it in promptly.

rotary/rules/polish.smk
jmtsuji: Thanks for bringing this up. The corresponding rule for the circularize module, `prepare_polypolish_circularize_input`, did not have `temp()` on its output file, so there was some inconsistency between the two rules. To address this, I decided to play things on the safe side and set the output of both `prepare_polypolish_circularize_input` and this rule to non-temp. I could try setting both to `temp()` instead but will need to test if that affects the pipeline running correctly as in #132 -- thoughts?

rotary/rules/polish.smk
jmtsuji: Indeed, pypolca cleans up its BWA files -- thanks for checking!

jmtsuji:
@LeeBergstrand I left comments for you, including a reply about the `temp()` designation of output files. Let me know your thoughts -- thanks.

rotary/rules/polish.smk
LeeBergstrand: @jmtsuji 

> I could try setting both to temp() instead but will need to test if that affects the pipeline running correctly as in https://github.com/rotary-genomics/rotary/issues/132 -- thoughts?

Let's try this.

LeeBergstrand:
> @LeeBergstrand I left comments for you, including a reply about the `temp()` designation of output files. Let me know your thoughts -- thanks.

@jmtsuji Let's try adding temp to both rules. I don't remember trying to add temp to one or the other on purpose. I think I intended to add temp to both.

rotary/rules/polish.smk
jmtsuji: Tests are done -- after changing the output of both rules to temp(), the behaviour seems identical to what was found for #132, so I think it's okay to set both as temp().

jmtsuji:
@LeeBergstrand Everything seems to be resolved (see the comments above) -- let me know if you notice anything else or if you are OK to merge this PR. If you are OK, then feel free to go ahead and merge this into develop!

LeeBergstrand:
Done!

PR #138, opened 2024-02-21
move keep_final_qc_read_files to true.

LeeBergstrand:
Move to using default `True` for keep_final_qc_read_files in config.yaml to address https://github.com/rotary-genomics/rotary/issues/132


jmtsuji (approved):
Thanks for making this change! Looks good to me.

PR #143, opened 2024-03-02
Make compatible with pungi updates

LeeBergstrand:
Changes to rotary to make it compatible with the latest pungi updates detailed here: https://github.com/rotary-genomics/pungi/pull/3

- move pathing to the default config back into rotary rather than pungi. Utilizes pungi's `get_config_path` function.
- pass the config object to `setup_run_directory` rather than the pass.
- some minor variable renaming and documentation improvements.


LeeBergstrand:
miving to dedicated dev branch for lee

PR #145, opened 2024-03-13
Develop lee

LeeBergstrand:
- various bug fixes to make the code compatible with the latest versions of pungi.
- change the default medaka model to "r1041_e82_400bps_hac_v4.2.0"
- add circularization TSV to the summary.zip

### DIFF
--- a/rotary/rotary.py
+++ b/rotary/rotary.py
@@ -123,7 +123,12 @@ def init(args):
 
     input_path = get_cli_arg_path(args, 'input_dir')
 
-    dataset = generate_dataset_from_fastq_directory(input_path, expected_files_per_sample=3)
+    if get_cli_arg(args, 'no_short'):
+        expected_files_per_sample = 1
+    else:
+        expected_files_per_sample = 3
+
+    dataset = generate_dataset_from_fastq_directory(input_path, expected_files_per_sample=expected_files_per_sample)
 
     dataset.create_sample_tsv(output_dir_path, header=sample_tsv_header_fields)
 
@@ -189,6 +194,8 @@ def parse_cli():
                              help='path to a directory containing Oxford Nanopore long-read and Illumina short-read .fastq(.gz) files')
     parser_init.add_argument('-f', '--force', action='store_true',
                              help="override existing run configuration files.")
+    parser_init.add_argument('-ns', '--no_short', action='store_true',
+                             help="do not search for short read files")
     parser_init.set_defaults(init=True)
     return parser
 
--- a/rotary/rotary.py
+++ b/rotary/rotary.py
@@ -10,7 +10,7 @@ import os
 
 from pungi.dataset import generate_dataset_from_fastq_directory, Dataset
 from pungi.run import setup_run_directory, run_snakemake_workflow, load_yaml_config, get_snakemake_args
-from pungi.sample import SequencingFile, LongReadSampleWithPairedPolishingShortReads
+from pungi.sample import SequencingFile, auto_create_sample_from_files
 from pungi.utils import get_cli_arg_path, get_cli_arg, check_for_files, get_config_path
 
 rotary_config_name = 'config.yaml'
@@ -86,17 +86,18 @@ def run_one(args):
     jobs = get_cli_arg(args, 'jobs')
     snakemake_args = get_snakemake_args(args)
 
-    sequencing_files = [SequencingFile(path) for path in [(get_cli_arg_path(args, 'long')),
-                                                          (get_cli_arg_path(args, 'left')),
-                                                          (get_cli_arg_path(args, 'right'))]]
+    sequencing_file_cli_paths = [get_cli_arg_path(args, file_type) for file_type in ['long', 'left', 'right']]
+    sequencing_files = [SequencingFile(path) for path in sequencing_file_cli_paths if path]
 
-    sample = LongReadSampleWithPairedPolishingShortReads(long_file=sequencing_files[0],
-                                                         short_file_left=sequencing_files[1],
-                                                         short_file_right=sequencing_files[2],
-                                                         # Don't do an identifier check on user-specified files.
-                                                         identifier_check=False,
-                                                         # Don't do integrity check on user-specified files.
-                                                         integrity_check=False)
+    if len(sequencing_files) in [1, 3]:
+        sample = auto_create_sample_from_files(*sequencing_files,
+                                               # Don't do an identifier check on user-specified files.
+                                               identifier_check=False,
+                                               # Don't do integrity check on user-specified files.
+                                               integrity_check=False)
+    else:
+        raise ValueError('There should either be one input sequencing file (long-read only) or three (hybrid '
+                         'sequencing)')
 
     dataset = Dataset(sample)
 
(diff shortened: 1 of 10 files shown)


rotary/rotary.py
```diff
@@ -60,17 +61,13 @@ def run(args):
...
-    config_path = get_cli_arg_path(args, 'config')
-    if not config_path:
-        config_path = os.path.join(output_dir_path, 'config.yaml')
+    config_path = get_config_path(args, default_config_path=os.path.join(output_dir_path, 'config.yaml'))
 
     config = load_yaml_config(config_path)
-
-    conda_env_directory = os.path.join(config['db_dir'], 'rotary_conda_env')
-    os.makedirs(conda_env_directory, exist_ok=True)
+    conda_env_dir = os.path.join(config['db_dir'], 'rotary_conda_envs')
```
jmtsuji, line 67: Sounds good to change the default conda env dir name.

rotary/rotary.py
```diff
@@ -112,16 +114,17 @@ def init(args):
...
     output_dir_path = get_cli_arg_path(args, 'output_dir')
 
     # Checks for existing run files and sets up the run directory.
-    setup_run_directory(args, output_dir_path, run_files)
+    setup_run_directory(args, output_dir_path, run_files, config)
 
     input_path = get_cli_arg_path(args, 'input_dir')
 
-    dataset = generate_dataset_from_fastq_directory(input_path)
+    dataset = generate_dataset_from_fastq_directory(input_path, expected_files_per_sample=3)
```
jmtsuji: Just adding a note: we'll need to generalize this eventually when adding compatibility for long read-only samples.

rotary/rules/annotation.smk
```diff
@@ -11,10 +12,7 @@ VERSION_GTDB_MAIN=VERSION_GTDB_COMPLETE.split('.')[0] # Remove subversion
 
 DB_DIR_PATH = config.get('db_dir')
 
-if str(config.get('keep_final_coverage_bam_files')).lower() == 'true':
-    KEEP_BAM_FILES = True
-else:
-    KEEP_BAM_FILES = False
+KEEP_BAM_FILES = is_config_parameter_true(config,'keep_final_coverage_bam_files')
```
jmtsuji, line 15: Nice simplification of the code :-)

rotary/rules/assembly.smk
```diff
@@ -104,9 +104,20 @@ rule finalize_assembly:
         os.symlink(source_relpath,str(output.info))
 
 
+rule add_circular_info_to_annotation_stats:
+    input:
+        "{sample}/assembly/{sample}_circular_info.tsv"
```
jmtsuji: You might consider changing this to `"{sample}/polish/{sample}_contig_info.tsv"`, which contains the circularization info from the `"{sample}/assembly/{sample}_circular_info.tsv"` file along with additional stats.

rotary/rules/assembly.smk
```diff
@@ -104,9 +104,20 @@ rule finalize_assembly:
...
+        """
+        cp {input} {output}
+        """
+
 rule assembly:
     input:
         expand("{sample}/assembly/{sample}_assembly.fasta",sample=SAMPLE_NAMES),
-        expand("{sample}/assembly/{sample}_circular_info.tsv", sample=SAMPLE_NAMES)
+        expand("{sample}/assembly/{sample}_circular_info.tsv", sample=SAMPLE_NAMES),
+        expand("{sample}/stats/{sample}_circular_info.tsv", sample=SAMPLE_NAMES)
```
jmtsuji: If the above reference is changed to `"{sample}/assembly/{sample}_circular_info.tsv"`, then this input should be changed as well (and moved from `rule assembly` to `rule polish`. What do you think?

rotary/rules/assembly.smk
jmtsuji: If this change is made, then this rule should be moved to `polish.smk`.

jmtsuji (changes requested):
Thanks for making these improvements! I left some very minor comments. Happy to discuss further about the stats file in particular. I think it should be straightforward to merge this PR.

rotary/rotary.py
LeeBergstrand: Resolved by: https://github.com/rotary-genomics/rotary/pull/146

jmtsuji: Great!

rotary/rules/assembly.smk
LeeBergstrand: @jmtsuji I'm not quite sure I understand. Can you make a pull request for the changes you want? I intended to add circularization info to `summary.zip` generated by the `summarize_annotation` rule in `annotation.smk`. I noticed that there wasn't a file in `summary.zip` that would tell you what contigs were linear or circular, which is valuable info for passing to end users.

LeeBergstrand:
@jmtsuji Left a comment above. Not sure of your intent in renaming `"{sample}/assembly/{sample}_circular_info.tsv"`.

rotary/rules/assembly.smk
jmtsuji: @LeeBergstrand Sorry that my comment wasn't clear! (Was writing quickly before trying to catch the bus...) Yes, I agree that stats info about the circular vs. linear status of the contigs should be added to the final summary.zip file. The `"{sample}/assembly/{sample}_circular_info.tsv"` file includes this information, for example:

```
contig    circular  repaired
contig_6  Y         Y
contig_4  Y         Y
contig_2  Y         Y
contig_5  Y         Y
contig_3  Y         N
contig_7  N         N
contig_1  N         N
```

However, the `"{sample}/assembly/{sample}_circular_info.tsv"` adds to the `"{sample}/polish/{sample}_contig_info.tsv"` file by including additional stats about whether the contigs passed the coverage filter, for example:

```
contig    circular  repaired  pass_coverage_filter
contig_6  Y         Y         Y
contig_4  Y         Y         Y
contig_2  Y         Y         Y
contig_5  Y         Y         Y
contig_3  Y         N         Y
contig_7  N         N         N
contig_1  N         N         N
```

So I think that the `"{sample}/polish/{sample}_contig_info.tsv"` might be a more suitable summary stats file than the `"{sample}/assembly/{sample}_circular_info.tsv"` file.

Eventually, I hope to make a similar stats file in the `circularization` folder that additionally includes whether or not the contigs could be rotated to a start gene. The concept of these log files is that they report the summary stats of all modules run up to that point for each contig. I also hope to eventually add more info, e.g., the contig length. But for now, I think the `"{sample}/polish/{sample}_contig_info.tsv"` file is the most informative contig log output by the current version of rotary.

jmtsuji: Does that help clarify? I can make a simple PR for this later today.

jmtsuji:
> @jmtsuji Left a comment above. Not sure of your intent in renaming `"{sample}/assembly/{sample}_circular_info.tsv"`.

@LeeBergstrand Just replied above -- thanks.

rotary/rules/assembly.smk
LeeBergstrand: @jmtsuji great! Ya this contig file is better thanks for the change.

jmtsuji: Great!

LeeBergstrand:
@jmtsuji Are we ready to merge this in?

jmtsuji (approved):
@LeeBergstrand Yes, OK to merge -- thanks so much for your work on these improvements!!

PR #146, opened 2024-04-03
Allow for no short read input

LeeBergstrand:
### Modify the code so that short reads are no longer required.

Previously, you would have to make mock short read files to get the pipeline to start if you didn't have short reads.

## Changes

- Modify CLI
  - modify `rotary run_one` so the `-r1` and `-r2` flags are optional.
  - modify `rotary init` to add a '--no_short` flag to tell the `rotary` that the input directory contains no short reads.
  - modify `rotary` and `pungi` codes so they can read and write a `samples.tsv` file with only a long sample file column but no short read file columns.
  -  modify `rotary.smk`'s `set_up_sample_directories` rule to take short reads optionally. Also modify it so if there are no short reads that the `POLISH_WITH_SHORT_READS` parameter is set to `False`

## Limitations

Sample directories can either contain only long and short reads or all long reads, but not a mix of both. The  `rotary init` 's `--no_short` cannot be activated while importing a directory containing short reads.

### DIFF
--- a/rotary/rotary.py
+++ b/rotary/rotary.py
@@ -123,7 +123,12 @@ def init(args):
 
     input_path = get_cli_arg_path(args, 'input_dir')
 
-    dataset = generate_dataset_from_fastq_directory(input_path, expected_files_per_sample=3)
+    if get_cli_arg(args, 'no_short'):
+        expected_files_per_sample = 1
+    else:
+        expected_files_per_sample = 3
+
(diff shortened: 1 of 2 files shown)


LeeBergstrand:
@jmtsuji This fixes https://github.com/rotary-genomics/rotary/issues/118

jmtsuji (approved):
@LeeBergstrand Thanks for this enhancement! Like you mentioned, this knocks out some of the "low-hanging fruit" for the longer term goal of #59. The code looks good to me -- feel free to merge.

Issue #131, opened 2024-02-15
Better parallelization for medaka


jmtsuji:
A possible way to speed up rotary: according to the [medaka GitHub repo](https://github.com/nanoporetech/medaka?tab=readme-ov-file#improving-parallelism), the consensus step of medaka is basically capped at 2 threads. It might make sense for us to implement the more advanced usage of medaka, as shown in the link above, to achieve maximum parallelization. Basically, we can use the `--regions` flag of `medaka consensus` to run different regions of the input BAM file (e.g., different contigs, or 1 Mb chunks of contigs) through the consensus step in parallel.

If we wanted to add this speed-up to rotary, we could either implement it directly in snakemake or by making a Python script that runs the medaka (including the consensus step) in parallel.

I don't think this is a high priority, but I wanted to post it here while it is on my mind.

LeeBergstrand:
@jmtsuji We should profile the pipeline first to figure out the slowest steps and target those first. Medaka is one of the slower steps, but I don't think it's the slowest.

jmtsuji:
Yes, sounds good to profile the pipeline and figure out what steps are the slowest. Let's table this for now.

LeeBergstrand:
@jmtsuji We now have polishing by contig in https://github.com/rotary-genomics/rotary/pull/147

LeeBergstrand:
Addressed in https://github.com/rotary-genomics/rotary/pull/147

If we run into speed issues, we can look into doing sub-contig polishing, but that might be an over optimization.

PR #147, opened 2024-04-06
Optimize medaka parallelization

LeeBergstrand:
@jmtsuji This is a bodge that partially addresses https://github.com/rotary-genomics/rotary/issues/131. After some testing it only makes the code 1% faster (code goes to 2 threads for the longest contig in the end). If you had a genome with lots of big contigs (multiple chromosomes) or multiple genomes then there would be greater speed ups.

@jmtsuji This code works. Can you review it to make sure I'm not making any mistakes in the overall logic?

Right now I am work on another branch https://github.com/rotary-genomics/rotary/tree/optimize_medaka_parallelization_by_contig to make it so it polishes each contig in a separate rule instanced spawned by a checkpoint reevaluation of the DAG after it figures out how what contigs were produced by the assembly. This is cleaner and more scalable (the schedular can interweave polishing across contigs and organisms if each polishing rule is set up as taking two threads) but a little harder to right code for.

--------

optimize_medaka_parallelization_by_contig has been merged in.

The code now uses a checkpoint and aggregating function to dynamically run medaka polishing by contig.

This should provide significant speed-ups when running multiple genomes or large chromosomes simultaneously.

### DIFF
--- a/rotary/envs/medaka.yaml
+++ b/rotary/envs/medaka.yaml
@@ -5,4 +5,4 @@ channels:
 dependencies:
   - medaka=1.8.0
   - parallel
-  - seqtk=1.3
+  - seqkit=2.8.0
--- a/rotary/envs/medaka.yaml
+++ b/rotary/envs/medaka.yaml
@@ -4,3 +4,5 @@ channels:
   - defaults
 dependencies:
   - medaka=1.8.0
+  - parallel
+  - seqtk=1.3
--- a/rotary/envs/medaka.yaml
+++ b/rotary/envs/medaka.yaml
@@ -4,4 +4,3 @@ channels:
   - defaults
 dependencies:
   - medaka=1.8.0
-  - seqkit=2.8.0
--- a/rotary/envs/medaka.yaml
+++ b/rotary/envs/medaka.yaml
@@ -4,5 +4,4 @@ channels:
   - defaults
 dependencies:
   - medaka=1.8.0
-  - parallel
   - seqkit=2.8.0
--- a/rotary/rules/circularize.smk
+++ b/rotary/rules/circularize.smk
@@ -272,14 +272,19 @@ rule bypass_circularization:
         os.symlink(source_relpath,str(output))
 
 
-# Gets the names of all lists in circularize/filter/lists (should be either circular.list or linear.list or both)
-# Then outputs the expected paths of the finalized circular and linear FastA files
-# This function allows the DAG to figure out whether to run the circular / linear specific processing steps
-#   based on the split_circular_and_linear_contigs checkpoint made earlier.
+
 def aggregate_contigs(wildcards):
-    # TODO - I do not further use this variable, but checkpoints needs to be called to trigger the checkpoint.
-    # Am I doing something wrong?
-    checkpoint_output = checkpoints.split_circular_and_linear_contigs.get(**wildcards).output[0]
+    """
+    Gets the names of all lists in circularize/filter/lists (should be either circular.list or linear.list or both)
+    Then outputs the expected paths of the finalized circular and linear FastA files
+    This function allows the DAG to figure out whether to run the circular / linear specific processing steps
+    based on the split_circular_and_linear_contigs checkpoint made earlier.
+    """
+    # Force execution of checkpoint split_circular_and_linear_contigs.
+    # Passes wildcards from rule combine_circular_and_linear_contigs to split_circular_and_linear_contigs.
+    # Execution will generate a lists of linear and circular contigs and revaluate the DAG.
+    checkpoints.split_circular_and_linear_contigs.get(**wildcards)
+
     circularize_lists_path = f"{wildcards.sample}/circularize/filter/lists"
 
     return expand("{{sample}}/circularize/combine/{{sample}}_{circular_or_linear}.fasta",
--- a/rotary/rules/polish.smk
+++ b/rotary/rules/polish.smk
@@ -42,11 +42,12 @@ rule map_long_reads_for_medaka:
         mini_align -i {input.qc_long_reads} -r {input.contigs} -m -p {params.output_base_name} -t {threads} > {log} 2>&1
         """
 
-checkpoint generate_contig_files:
+checkpoint generate_contig_manifest:
     input:
         "{sample}/{step}/medaka_input/{sample}_input.fasta"
     output:
-        directory("{sample}/{step}/medaka/results")
+        contig_manifest="{sample}/{step}/medaka/results/{sample}_contigs.txt",
+        medaka_results_dir=directory('{sample}/{step}/medaka/results')
     conda:
         "../envs/medaka.yaml"
     log:
@@ -55,8 +56,7 @@ checkpoint generate_contig_files:
         "{sample}/benchmarks/{step}/medaka.txt"
     shell:
         """
-        mkdir -p {output}
-        seqkit seq -n {input} | parallel 'touch {output}/{wildcards.sample}_{{}}' 2>> {log}
+        seqkit seq -n {input} > {output.contig_manifest} 2>> {log}
         """
 
 rule polish_contig_medaka:
@@ -64,7 +64,7 @@ rule polish_contig_medaka:
         calls_to_draft_bam='{sample}/{step}/medaka/calls_to_draft.bam',
         calls_to_draft_bam_index='{sample}/{step}/medaka/calls_to_draft.bam.bai',
         medaka_results_dir='{sample}/{step}/medaka/results',
-        contig_name='{sample}/{step}/medaka/results/{sample}_{contig}'
+        contig_manifest="{sample}/{step}/medaka/results/{sample}_contigs.txt"
     output:
         contig_polished='{sample}/{step}/medaka/results/{sample}_{contig}.hd5'
     conda:
@@ -95,15 +95,15 @@ def aggregate_medaka_polished_contigs(wildcards):
     :param wildcards: These are the wildcards present in rule stich_medaka.
     :return: HDF5 files to be generated for each contig by multiple executions of rule polish_contig_medaka.
     """
-    # Force execution of checkpoint generate_contig_files. This should take the input assembly,
-    # generate a series of stub files representing each contig and then revaluate the DAG.
-    # Wildcards from rule stitch_medaka are passed to checkpoint generate_contig_files.
-    checkpoints.generate_contig_files.get(**wildcards)
+    # Force execution of checkpoint generate_contig_files.
+    # This will generate a contig manifest file and revaluate the DAG.
+    contig_manifest_path = checkpoints.generate_contig_manifest.get(**wildcards).output[0]
 
-    # Uses the names of the stub files to get the names of the contig files.
-    contigs_names = glob_wildcards(f"{wildcards.sample}/{wildcards.step}/medaka/results/{wildcards.sample}_{{contig}}").contig
+    # Open the resulting contig manifest file and read in the contig names.
+    with contig_manifest_path.open() as contig_file:
+        contigs_names = [line.strip() for line in contig_file]
 
-    # Returns the expected paths to the per-contig HDF5 files.
+    # Returns the expected paths to the per-contig HDF5 polishing files.
     return expand("{sample}/{step}/medaka/results/{sample}_{contig}.hd5",sample=wildcards.sample,
         step=wildcards.step,contig=contigs_names)
 
--- a/rotary/rules/polish.smk
+++ b/rotary/rules/polish.smk
@@ -47,16 +47,14 @@ checkpoint generate_contig_manifest:
     input:
         "{sample}/{step}/medaka_input/{sample}_input.fasta"
     output:
-        contig_manifest="{sample}/{step}/medaka/{sample}_contigs.txt"
-    conda:
-        "../envs/medaka.yaml"
+        "{sample}/{step}/medaka/{sample}_contigs.txt"
     log:
         "{sample}/logs/{step}/contig_names_medaka.log"
     benchmark:
         "{sample}/benchmarks/{step}/medaka.txt"
     shell:
         """
-        seqkit seq -n {input} > {output.contig_manifest} 2>> {log}
+        grep '^>' {input} | cut -f1 -d' ' | tr -d '>' > {output} 2>> {log}
         """
 
 rule polish_contig_medaka:
--- a/rotary/rules/polish.smk
+++ b/rotary/rules/polish.smk
@@ -62,8 +62,7 @@ checkpoint generate_contig_manifest:
 rule polish_contig_medaka:
     input:
         calls_to_draft_bam='{sample}/{step}/medaka/calls_to_draft.bam',
-        calls_to_draft_bam_index='{sample}/{step}/medaka/calls_to_draft.bam.bai',
-        contig_manifest="{sample}/{step}/medaka/{sample}_contigs.txt"
+        calls_to_draft_bam_index='{sample}/{step}/medaka/calls_to_draft.bam.bai'
     output:
         contig_polished=temp('{sample}/{step}/medaka/results/{sample}_{contig}.hd5')
     conda:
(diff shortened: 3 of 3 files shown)


LeeBergstrand:
@jmtsuji Added checkpoint-based code for running medaka polish per contig.

LeeBergstrand:
@jmtsuji Update code to use a text file containing the names of the contigs.

LeeBergstrand:
@jmtsuji, https://github.com/rotary-genomics/rotary/issues/131 should be fully addressed. I'm looking forward to your review. I haven't benchmarked it yet, but it feels like Medaka now goes much faster with input sample multiple genomes.

LeeBergstrand:
@jmtsuji I also wrote better documentation and cleaned up the function `aggregate_contigs` in `circularize.smk`. You might need to document it to a similar level to `aggregate_medaka_polished_contigs` in `polish.smk`. Thoughts?

rotary/rules/polish.smk
```diff
@@ -15,37 +15,120 @@ rule prepare_medaka_polish_input:
...
+        contig_manifest="{sample}/{step}/medaka/{sample}_contigs.txt"
+    conda:
+        "../envs/medaka.yaml"
+    log:
+        "{sample}/logs/{step}/contig_names_medaka.log"
     benchmark:
         "{sample}/benchmarks/{step}/medaka.txt"
+    shell:
+        """
+        seqkit seq -n {input} > {output.contig_manifest} 2>> {log}
```
jmtsuji: @LeeBergstrand Does this just create a list of contig names? If so, I wonder if we might be able to save a dependency and do this using something like grep, awk, or sed?

rotary/envs/medaka.yaml
```diff
@@ -4,3 +4,5 @@ channels:
   - defaults
 dependencies:
   - medaka=1.8.0
+  - parallel
```
jmtsuji: Is `parallel` no longer needed in the final code version, given that aggregation is handled by snakemake? Just checking.

rotary/rules/polish.smk
```diff
@@ -15,37 +15,120 @@ rule prepare_medaka_polish_input:
...
+        """
+
+rule polish_contig_medaka:
+    input:
+        calls_to_draft_bam='{sample}/{step}/medaka/calls_to_draft.bam',
+        calls_to_draft_bam_index='{sample}/{step}/medaka/calls_to_draft.bam.bai',
+        contig_manifest="{sample}/{step}/medaka/{sample}_contigs.txt"
```
jmtsuji: It seems like this input (`contig_manifest`) is not called in the shell section below -- is that correct?

rotary/rules/polish.smk
```diff
@@ -15,37 +15,120 @@ rule prepare_medaka_polish_input:
...
         medaka_model=config.get("medaka_model"),
-        batch_size=config.get("medaka_batch_size")
+        batch_size=config.get("medaka_batch_size"),
+        medaka_results_dir='{sample}/{step}/medaka/results',
+    threads:
+        2
+    shell:
+        """
+        medaka consensus {input.calls_to_draft_bam} \
+          {params.medaka_results_dir}/{wildcards.sample}_{wildcards.contig}.hd5 \
```
jmtsuji: Does this line need to be converted into an output variable to be recognized by the DAG (e.g., so the files can be deleted if the rule fails, etc.)?

rotary/rules/polish.smk
```diff
@@ -15,37 +15,120 @@ rule prepare_medaka_polish_input:
...
+        contigs_names = [line.strip() for line in contig_file]
+
+    # Returns the expected paths to the per-contig HDF5 polishing files.
+    return expand("{sample}/{step}/medaka/results/{sample}_{contig}.hd5",sample=wildcards.sample,
+        step=wildcards.step,contig=contigs_names)
+
+rule stitch_medaka:
+    input:
+        hdf5s=aggregate_medaka_polished_contigs,
+        draft_fasta_dir="{sample}/polish/medaka_input/",
```
jmtsuji, line 109: Just checking: is this directory input variable redundant now? It seems like it is not called in the shell section, and I couldn't think of another reason why it might need to be there.

rotary/rules/polish.smk
```diff
@@ -15,37 +15,120 @@ rule prepare_medaka_polish_input:
...
         config.get("threads",1)
+    params:
+        medaka_model=config.get("medaka_model"),
+        batch_size=config.get("medaka_batch_size")
     shell:
         """
-        medaka_consensus -i {input.qc_long_reads} -d {input.contigs} -o {output.dir} \
-          -m {params.medaka_model} -t {threads} -b {params.batch_size} > {log} 2>&1
-        mv {output.dir}/consensus.fasta {output.contigs}
+        medaka stitch --threads {threads} {input.hdf5s} {input.draft_fasta} {output} > {log} 2>&1
```
jmtsuji, line 126: I think this is an inherent issue with the medaka stitch CLI itself, but note that, if the assembly file contains a huge number of contigs, the shell will error out with an "argument list too long" error due to the large number of HDF5 files (e.g., see https://stackoverflow.com/a/19355351)

rotary/rules/polish.smk
```diff
@@ -15,37 +15,120 @@ rule prepare_medaka_polish_input:
...
+    # Passes wildcards from rule stitch_medaka to generate_contig_files.
+    # Execution will generate a contig manifest file and revaluate the DAG.
+    contig_manifest_path = checkpoints.generate_contig_manifest.get(**wildcards).output[0]
+
+    # Open the resulting contig manifest file and read in the contig names.
+    with contig_manifest_path.open() as contig_file:
+        contigs_names = [line.strip() for line in contig_file]
+
+    # Returns the expected paths to the per-contig HDF5 polishing files.
+    return expand("{sample}/{step}/medaka/results/{sample}_{contig}.hd5",sample=wildcards.sample,
```
jmtsuji, line 103: This is a nice solution! Great that it is generalized by `step` as well.

rotary/rules/polish.smk
```diff
@@ -15,37 +15,120 @@ rule prepare_medaka_polish_input:
...
+        "../envs/medaka.yaml"
+    log:
+        "{sample}/logs/{step}/medaka_{sample}_{contig}.log"
+    benchmark:
+        "{sample}/benchmarks/{step}/medaka_{sample}_{contig}.txt"
     params:
         medaka_model=config.get("medaka_model"),
-        batch_size=config.get("medaka_batch_size")
+        batch_size=config.get("medaka_batch_size"),
+        medaka_results_dir='{sample}/{step}/medaka/results',
```
jmtsuji: If the line flagged below (84) is converted into an output variable, then I wonder if this param can be deleted?

rotary/rules/circularize.smk
```diff
@@ -272,14 +272,19 @@ rule bypass_circularization:
...
-# Then outputs the expected paths of the finalized circular and linear FastA files
-# This function allows the DAG to figure out whether to run the circular / linear specific processing steps
-#   based on the split_circular_and_linear_contigs checkpoint made earlier.
+
 def aggregate_contigs(wildcards):
-    # TODO - I do not further use this variable, but checkpoints needs to be called to trigger the checkpoint.
-    # Am I doing something wrong?
-    checkpoint_output = checkpoints.split_circular_and_linear_contigs.get(**wildcards).output[0]
+    """
+    Gets the names of all lists in circularize/filter/lists (should be either circular.list or linear.list or both)
```
jmtsuji, line 278: Thanks very much for this initial cleanup of the documentation for this aggregation function! Very helpful. I'll take a look at this during future code dev to see if I can clarify the documentation further.

rotary/rules/polish.smk
jmtsuji: Or does it need to be outside the output section to work properly with checkpointing somehow?

jmtsuji (changes requested):
@LeeBergstrand Thanks very much for adding this enhancement to parallelization! The overall logic looks good to me, although I did not test the code to see if it works in practice (it sounds like you've tested it successfully?). I wonder if some parts of the code are leftover from the intermediate version when GNU Parallel was used, so I flagged a couple things that seemed like they might be extraneous. Exciting that it sounds like you've seen good speedups with multiple genomes. This should also be a great addition for doing metagenome assembly, like we chatted about. Let me know your feedback about my comments, and then I think we should be good to merge!

rotary/rules/polish.smk
LeeBergstrand: Yes, this dumps a list of contig identifiers to a file. 

Over the years, I have generally leaned towards using a dedicated bioinformatics tool rather than Unix tools to do header extraction because I have found that people are very liberal with the text of their faster headers. For example, people put `-->` in their fasta headers and other weird characters, or they put spaces or they put "(" and ")". All these things tend to break the parsers you write yourself. 

If you use a dedicated tool, the developer has coded in logic that sanitizes the header identifiers by stripping out bad characters or other info, which makes things stabler.

LeeBergstrand: However, because this step will only get fasta files generated by an assembler as input we might be able to go to grep, awk, or sed. What are your thoughts?

rotary/rules/polish.smk
LeeBergstrand: So the long-read mapping generates `.mmi` and `.fai` files in the same folder as the draft FASTA (e.g., `{sample}/polish/medaka_input/{sample}_input.fasta.mmi`).

```python
rule prepare_medaka_polish_input:
    input:
        "{sample}/assembly/{sample}_assembly.fasta"
    output:
        fasta=temp("{sample}/polish/medaka_input/{sample}_input.fasta"),
        fasta_dir=temp(directory("{sample}/polish/medaka_input/"))
```

In `prepare_medaka_polish_input,` the draft FASTA was initially set as a temp(). So, the FASTA would be deleted but not the `mmi` and `fai` files generated in that folder by `map_long_reads_for_medaka.` One can't directly reference the `fai` and `mmi` files in `prepare_medaka_polish_input` because they haven't been created yet. You can't delete them in the `map_long_reads_for_medaka` because they don't have a `step` wild card. So I found a way to delete the entire `"{sample}/polish/medaka_input` directory by setting it to temp in `prepare_medaka_polish_input.` If you don't reference the directory later in the DAG, the entire folder will be deleted immediately after the rule runs, deleting the draft FASTA. So, I referenced the draft FASTA directory as input into `stitch_medaka`; this prevents the directory from being deleted until the polishing is complete.

LeeBergstrand: Does that work or do you think there is a better way to do it?

rotary/rules/polish.smk
LeeBergstrand: @jmtsuji How common do you think this will be? This would be a corner case, no? If you have many short contigs (high tens or hundreds), you have more significant issues with the nanopore run or need to change your trim settings. I suspect that you would be able to feed it a lot more HDF5 paths than we think.

LeeBergstrand: So, if you look at the code for `medaka_concensus,` it's a shell script that contains all the steps in the medaka pipeline. It calls `medaka stitchch,` as a CLI command, so it should run into the same issue.

https://github.com/nanoporetech/medaka/blob/master/scripts/medaka_consensus

LeeBergstrand: So I think we should leave this. But thanks for the heads up about a potential issue.

rotary/rules/polish.smk
LeeBergstrand: Addressed

rotary/rules/polish.smk
LeeBergstrand: resolved

rotary/rules/polish.smk
LeeBergstrand: resolved

rotary/rules/polish.smk
LeeBergstrand: @jmtsuji Swapped out for shell commands.

rotary/envs/medaka.yaml
LeeBergstrand: removed

rotary/rules/polish.smk
LeeBergstrand: @jmtsuji I think we can leave this agreed?

rotary/rules/circularize.smk
LeeBergstrand: Great

LeeBergstrand: made issue

LeeBergstrand:
> @LeeBergstrand Thanks very much for adding this enhancement to parallelization! The overall logic looks good to me, although I did not test the code to see if it works in practice (it sounds like you've tested it successfully?). I wonder if some parts of the code are leftover from the intermediate version when GNU Parallel was used, so I flagged a couple things that seemed like they might be extraneous. Exciting that it sounds like you've seen good speedups with multiple genomes. This should also be a great addition for doing metagenome assembly, like we chatted about. Let me know your feedback about my comments, and then I think we should be good to merge!

I have tested this successfully on both single and multiple samples.

rotary/rules/polish.smk
jmtsuji: @LeeBergstrand Thanks for the discussion! Yeah, that makes sense to try to be careful, in general, of weird things in FastA headers. Because we are working more-or-less directly with the output from Flye like you mentioned, I think the shell code you've added should be OK. I think the shell code might need a minor modification so that STDERR from all three commands go to the log file:

Current
```bash
grep '^>' {input} | cut -f1 -d' ' | tr -d '>' > {output} 2>> {log}
```

Modified (with a subshell):
```bash
(grep '^>' {input} | cut -f1 -d' ' | tr -d '>' > {output}) 2> {log}
```
Note that I also changed `2>>` to `2>` so that it will overwrite rather than append to an existing log.

Does that look okay to you? If so, I am supportive of switching to this shell-based approach... it might leave us with one less software version to keep updated in the long term. Hopefully the shell code is simple enough to be compatible across multiple platforms and not easily broken... I tested the modified code in both bash and sh, and it seems to work as expected.

rotary/rules/polish.smk
jmtsuji: Ah, I see! Thanks for the detailed explanation. Yes, the way you've written the code makes sense to keep the `medaka_input` dir as temp.

It might help to add brief a comment above rule `stitch_medaka` to clarify why the `medaka_input` dir is declared as an input, so that someone doesn't delete this input line during code cleanup in the future. Otherwise, I am OK to keep the code as-is!

rotary/rules/polish.smk
jmtsuji: @LeeBergstrand Thanks for the additional check of the `medaka_consensus` code -- interesting that this is a shell script! I agree with you, let's leave our code as-is for now.

The possible case I thought of when our code might break is if someone wanted to use our pipeline to assemble a soil metagenome or something. They might have hundreds of thousands of contigs. In the `medaka_consensus` shell script, I think all those contigs would be kept in one HDF5 file rather than being split up into 1 contig per HDF5, because that script is not parallelized (in my understanding). So our code might break in this case, whereas the `medaka_consensus` script might still work. But it's probably OK for us to face that issue later if/when it comes up. We might need to batch the HDF5's differently or stitch in several batches or something.

rotary/rules/polish.smk
```diff
@@ -47,23 +47,20 @@ checkpoint generate_contig_manifest:
...
-        "../envs/medaka.yaml"
+        "{sample}/{step}/medaka/{sample}_contigs.txt"
     log:
         "{sample}/logs/{step}/contig_names_medaka.log"
     benchmark:
         "{sample}/benchmarks/{step}/medaka.txt"
     shell:
         """
-        seqkit seq -n {input} > {output.contig_manifest} 2>> {log}
+        grep '^>' {input} | cut -f1 -d' ' | tr -d '>' > {output} 2>> {log}
```
jmtsuji, line 57: See my reply at https://github.com/rotary-genomics/rotary/pull/147#discussion_r1559003859

jmtsuji (changes requested):
Looks good to me, with one minor comment. After addressing my proposed changes to the shell code in the checkpoint rule, feel free to go ahead and merge in these changes (no need for another round of review, I think, unless something unexpected comes up). Thanks so much for your hard work on this helpful PR!

rotary/rules/polish.smk
LeeBergstrand: @jmtsuji I will make the change you requested to the shell code. Yes, I think we should go with the shell code.

rotary/rules/polish.smk
LeeBergstrand: Agreed

rotary/rules/polish.smk
LeeBergstrand: @jmtsuji Yes 1) we would have to evaluate how many contigs would cause this failure to occur. I suspect that it's probably much higher than expected. 2) We would probably have to add other steps and changes to support metagenomes explicitly. So we can think about that issue then.

jmtsuji: Okay, sounds good. Let's consider this more later, if/when needed.

Issue #140, opened 2024-02-22
Make annotation sub-rules optional


jmtsuji:
Currently, unless the `--until` snakemake flag is used, rotary will run all steps in the annotation module. At the moment, these steps include:
- ORF prediction and basic functional annotation by DFAST
- Functional annotation by EggNOG mapper
- Taxonomy assignment by GTDB-Tk
- Quality check by CheckM2
- Read mapping for coverage information

Some of these steps are quite time (or memory) consuming or need a lot of disk space for DB download, and these factors can make testing difficult. Also, some users might just want a finished genome without all the detailed annotations.

@LeeBergstrand What would you think about adding a section to the config file where the desired annotations can be specified? Something like:

```
# Annotation sub-rules for rotary to run (comment out or delete lines to skip annotation types)
annotations:
- functional_annotations_DFAST
- functional_annotation_EggNOG_mapper
- taxonomy_assignment_GTDBTk
- quality_check_CheckM2 # Must be set for automated quality check to run
- read_coverage
```

Regarding DFAST: I think we should always provide gene/ORF calls for users, but I think functional annotation by DFAST is something we could potentially make optional. I checked, and it is possible to run DFAST with a `--no_func_anno` flag to skip functional annotation. Thus, I am thinking that the `functional_annotations_DFAST` point could be used to toggle DFAST's `--no_func_anno` flag.

I think it might be fairly easy to implement this optional annotation concept. Most of the options could be implemented by just by adding some conditionals to the `summarize_annotation` rule. To implement the `--no_func_anno` version of DFAST, we might need to either add conditions to the `run_dfast` rule or make a second rule called something like `run_dfast_simple`. Not sure if the `--no_func_anno` version of DFAST would still need a DB dir... we would need to test this.

One benefit of adding optional annotation is that we would not need to worry as much about supporting `--until circularize` as a common use case. I am using `--until circularize` at the moment to bypass the annotation module.

@LeeBergstrand Let me know your thoughts about this optional annotation sub-rule idea -- thanks!

jmtsuji:
> To implement the --no_func_anno version of DFAST, we might need to either add conditions to the run_dfast rule or make a second rule called something like run_dfast_simple. Not sure if the --no_func_anno version of DFAST would still need a DB dir... we would need to test this.

[...]

PR #154, opened 2024-04-16
Allow annotation steps to be be turned on and off in config.

LeeBergstrand:
Addresses issue https://github.com/rotary-genomics/rotary/issues/140

Allows the user to select what annotations to run based on a list of annotation types in the config file.

```yaml
annotations: ['DFAST_Func', 'EggNOG', 'GTDBTk', 'CheckM2', 'coverage']
```

Changes:

1. A Python class for creating objects that can be queried to determine what annotation steps should run.
2. Code that uses this class to determine what annotations to run.
3. Pass the `--no_func_anno` flag to DFAST if you don't want the light annotation to occur.
4. Use `{sample}_genome.fna` from DFAST as an indicator to symlink the log files to go inside `annotation_summary.zip` as, at minimum, DFAST is always run, unlike the other annotation types.

@jmtsuji Thoughts?

### DIFF
--- a/rotary/annotation.py
+++ b/rotary/annotation.py
@@ -1,0 +1,48 @@
+#!/usr/bin/env python
+
+"""
+Created by: Lee Bergstrand (2024)
+
+Description: Code for dealing with genome annotations.
+"""
+
+
+class AnnotationMap(object):
+    """
+    A class representing a map of the genome annotation bioinformatics software to be run on the genome.
+    """
+
+    def __init__(self, annotations):
+        """
+        Initialize the object with the provided annotations.
+        Args:
+            annotations (list[str]): A list annotation software to be run.
+        Dynamic Attributes:
+            dfast_func (bool): True if 'dfast_func' annotation should be run, False otherwise.
+            eggnog (bool): True if 'eggnog' annotation should be run, False otherwise.
+            gtdbtk (bool): True if 'gtdbtk' annotation should be run, False otherwise.
+            checkm2 (bool): True if 'checkm2' annotation should be run, False otherwise.
+            coverage (bool): True if 'coverage' annotation should be run, False otherwise.
+        """
+        self.expected_annotations = ['dfast_func', 'eggnog', 'gtdbtk', 'checkm2', 'coverage']
+
+        # Convert the annotations to be lower cases in case the user makes a small typo.
+        annotations_lower = [anno.lower() for anno in annotations]
+
+        for current_annotation in self.expected_annotations:
+            if current_annotation in annotations_lower:
+                setattr(self, current_annotation, True)  # Set the attributes dynamically.
+            else:
+                setattr(self, current_annotation, False)
+
+    def __repr__(self):
+        """
+        Generate a representation of the object state in the form of AnnotationMap(dfast_func=..., eggnog=..., ...).
+        """
+        repr_parts = []
+        for current_annotation in self.expected_annotations:
+            repr_parts.append(f"{current_annotation}={getattr(self, current_annotation)}")
+        return f"AnnotationMap({', '.join(repr_parts)})"
+
+    def __str__(self):
+        return self.__repr__()
--- a/rotary/rules/annotation.smk
+++ b/rotary/rules/annotation.smk
@@ -3,6 +3,7 @@
 
 import os
 from pungi.utils import is_config_parameter_true
+from rotary.annotation import AnnotationMap
 
 VERSION_DFAST="1.2.18"
 VERSION_EGGNOG="5.0.0" # See http://eggnog5.embl.de/#/app/downloads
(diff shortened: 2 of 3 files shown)


rotary/rules/annotation.smk
```diff
@@ -172,7 +175,8 @@ rule run_dfast:
         "{sample}/benchmarks/annotation/annotation_dfast.txt"
     params:
         db=directory(os.path.join(DB_DIR_PATH,"dfast_" + VERSION_DFAST)),
-        strain='{sample}'
+        strain='{sample}',
+        no_functional_annotation="--no_func_anno" if not ANNOTATION_MAP.dfast_func else ''
```
jmtsuji: When I tested DFAST in `--no_func_anno` mode, I found that the DB directory (i.e., `--dbroot`) does not need to be specified in this mode. This is nice, because it means that a user could install rotary and use it in `--no_func_anno` mode without needing to download any large annotation DBs for DFAST. If possible, it would be great if we could add the ability to skip the `--dbroot` argument in the shell command when DFAST runs in `--no_func_anno`.

Required changes:
- Add an if-else statement to the `db` param similar to what is used for the `no_functional_annotation` param. If `dfast_func` isn't specified, then `''` is returned, but otherwise `--dbroot {path to the db}` could be returned. The shell command would need to be modified to replace `--dbroot {params.db}` with just `{params.db}`.
- Only require the `install_finished` input if `dfast_func` is specified.
  - For clarity, `install_finished` could be renamed `annotation_db_install_finished` or something similar
  
@LeeBergstrand This seems relatively straightforward to implement -- thoughts?

rotary/annotation.py
```diff
@@ -0,0 +1,48 @@
...
+        # Convert the annotations to be lower cases in case the user makes a small typo.
+        annotations_lower = [anno.lower() for anno in annotations]
+
+        for current_annotation in self.expected_annotations:
+            if current_annotation in annotations_lower:
+                setattr(self, current_annotation, True)  # Set the attributes dynamically.
+            else:
+                setattr(self, current_annotation, False)
+
+    def __repr__(self):
```
jmtsuji, line 38: Nice chance for me to study `repr` and `str` methods -- thanks :-)

rotary/rules/annotation.smk
```diff
@@ -376,11 +381,11 @@ rule symlink_logs:
...
-        "{sample}/annotation/eggnog/{sample}.emapper.annotations",
-        "{sample}/annotation/gtdbtk/{sample}_gtdbtk.summary.tsv",
-        "{sample}/annotation/checkm/",
-        expand("{{sample}}/annotation/coverage/{{sample}}_{type}_coverage.tsv",
-            type=["short_read", "long_read"] if POLISH_WITH_SHORT_READS == True else ["long_read"]),
+        "{sample}/annotation/eggnog/{sample}.emapper.annotations" if ANNOTATION_MAP.eggnog else [],
+        "{sample}/annotation/gtdbtk/{sample}_gtdbtk.summary.tsv" if ANNOTATION_MAP.gtdbtk else [],
+        "{sample}/annotation/checkm/" if ANNOTATION_MAP.checkm2 else [],
+        "{sample}/annotation/coverage/{sample}_short_read_coverage.tsv" if ANNOTATION_MAP.coverage else [],
+        "{sample}/annotation/coverage/{sample}_long_read_coverage.tsv" if POLISH_WITH_SHORT_READS and ANNOTATION_MAP.coverage else [],
```
jmtsuji: I think `POLISH_WITH_SHORT_READS` should be in the conditional for the `{sample}_short_read_coverage` (rather than the long read coverage) file, correct?

rotary/rules/annotation.smk
jmtsuji: Let me know if it would be easier for me to push a commit for this.

jmtsuji (changes requested):
@LeeBergstrand Nice enhancement -- thank you for working on this while I focus on utility development! In my review, I've flagged a couple minor things to consider enhancing or fixing, but otherwise, looks great. I'd be happy to take a second look before it's finally merged. Thanks!

rotary/rules/annotation.smk
LeeBergstrand: @jmtsuji Good catch. Swapped.

rotary/rules/annotation.smk
LeeBergstrand: @jmtsuji Addressed

LeeBergstrand: Good catch!

LeeBergstrand:
> @LeeBergstrand Nice enhancement -- thank you for working on this while I focus on utility development! In my review, I've flagged a couple minor things to consider enhancing or fixing, but otherwise, looks great. I'd be happy to take a second look before it's finally merged. Thanks!

@jmtsuji Great job with the feedback. Good bug catch and feature request. Requesting that you review it again.

jmtsuji (approved):
@LeeBergstrand Thanks for incorporating these additional changes. Looks great -- feel free to merge in!

PR #163, opened 2024-05-09
Install pungi and spokewrench via conda YAML files

jmtsuji:
Moves pungi and spokewrench installs into conda environment configuration (YAML) files. Now, these tools can be installed automatically rather than needing to be installed manually (following conda commands).


jmtsuji:
Merging this into the `rotary_utils` branch for review as part of #152

LeeBergstrand:
@jmtsuji Added a few minor updates to the readme and pyproject.toml and then merged in.

jmtsuji:
@LeeBergstrand Excellent, thanks!

PR #174, opened 2024-05-27
Change GTDB URL to use mirror site for much faster database downloads.

LeeBergstrand:
Could you change the GTDB URL to the mirror site instead of the primary download site?

@jmtsuji I found that downloading GTDB from the primary site (https://data.gtdb.ecogenomic.org/) is slow because the server is overloaded. Its download rate is currently kilobytes per second (May 2024). It would take multiple days to download this data at that speed. I changed the GTDB URL to use their mirror site (https://data.ace.uq.edu.au/public/gtdb/data/releases/) for order of magnitude faster GTDB database download speeds.


LeeBergstrand:
@jmtsuji If this becomes a persistent issue, then we can perhaps add the GTDB data to [Zenodo](https://zenodo.org/)

jmtsuji (changes requested):
@LeeBergstrand Sounds fine to me to use the GTDB's mirror site for downloads. That's good to know that you got much faster download speeds on the mirror site. I downloaded the GTDB from the main site shortly after the new release came out, and like you mentioned, it took 3-4 days to download. (I had hoped that the connection would speed up again after a week or two, but I guess they are still getting heavy demand or are having other network issues.) Hopefully we won't need to pursue the Zenodo route, but we can consider that in future if needed.

I've reviewed the code, and everything looks good. 👍 My only proposed change is to merge these commits into `add_assembly_stats` rather than `develop`, given that other changes from `add_assembly_stats` seem to be included in this PR. If you agree, feel free to merge this PR into `add_assembly_stats`!

PR #175, opened 2024-05-28
Add download command

LeeBergstrand:
Add code to enable a Rotary download command to pre-download all databases required by rotary and prepare them for later use.

Addresses issue https://github.com/rotary-genomics/rotary/issues/45

### DIFF
--- a/rotary/rotary.py
+++ b/rotary/rotary.py
@@ -158,6 +158,10 @@ def download(args):
     jobs = get_cli_arg(args, 'jobs')
     snakemake_args = get_snakemake_args(args)
 
+    # If there are no snakemake args. Create them so they can be appended to.
+    if not snakemake_args:
+        snakemake_args = []
+
     # Tell snakemake to only run the download rule.
     snakemake_args.append('download')
 
--- a/rotary/rotary.py
+++ b/rotary/rotary.py
@@ -213,7 +213,7 @@ def parse_cli():
     parser_run_one.add_argument('-o', '--output_dir', metavar='PATH', default=os.getcwd(),
                                 help='path the output/rotary project directory')
     parser_run_one.add_argument('-d', '--database_dir', metavar='PATH',
-                                help='path the rotary database directory')
+                                help='path to the rotary database directory')
     parser_run_one.add_argument('-j', '--jobs', metavar='JOBS',
                                 help='number of threads that rotary should use (overrides config)')
     parser_run_one.add_argument('-s', '--snakemake_args', metavar='',
@@ -230,7 +230,7 @@ def parse_cli():
     parser_init.add_argument('-o', '--output_dir', metavar='PATH',
                              help='path the output/rotary project directory', default=os.getcwd())
     parser_init.add_argument('-d', '--database_dir', metavar='PATH', required=True,
-                             help='path the rotary database directory')
+                             help='path to the rotary database directory')
     parser_init.add_argument('-i', '--input_dir', metavar='PATH', required=True,
                              help='path to a directory containing Oxford Nanopore long-read and Illumina short-read .fastq(.gz) files')
     parser_init.add_argument('-f', '--force', action='store_true',
@@ -244,7 +244,7 @@ def parse_cli():
     download_help = """Downloads databases required by rotary to a specific directory """
     parser_download = subparsers.add_parser('download', help=download_help)
     parser_download.add_argument('-d', '--database_dir', metavar='PATH', required=True,
-                                 help='path the rotary database directory')
+                                 help='path to the rotary database directory')
     parser_download.add_argument('-c', '--config', metavar='YAML',
                                  help='path to the rotary yaml config file')
     parser_download.add_argument('-j', '--jobs', metavar='JOBS',
--- a/rotary/rotary.py
+++ b/rotary/rotary.py
@@ -10,7 +10,8 @@ import os
 
 from pungi.dataset import generate_dataset_from_fastq_directory, Dataset
 from pungi.run import setup_run_directory, run_snakemake_workflow, load_yaml_config, get_snakemake_args, \
-    validate_yaml_config_against_default
+    validate_yaml_config_against_default, modify_config_with_available_computational_resources, setup_database_dir, \
+    dump_yaml_config
 from pungi.sample import SequencingFile, auto_create_sample_from_files
 from pungi.utils import get_cli_arg_path, get_cli_arg, check_for_files, get_config_path
 
@@ -42,6 +43,8 @@ def main():
         run_one(args)
     elif hasattr(args, 'init'):
         init(args)
+    elif hasattr(args, 'download'):
+        download(args)
     else:
         parser.print_help()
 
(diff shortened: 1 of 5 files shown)


LeeBergstrand:
@jmtsuji Tested with dry run be feel free to review.

rotary/rotary.py
```diff
@@ -200,6 +238,21 @@ def parse_cli():
     parser_init.add_argument('-ns', '--no_short', action='store_true',
                              help="do not search for short read files")
     parser_init.set_defaults(init=True)
+
+    # =========================
+    # Declare Download Sub-command
+    download_help = """Downloads databases required by rotary to a specific directory """
+    parser_download = subparsers.add_parser('download', help=download_help)
+    parser_download.add_argument('-d', '--database_dir', metavar='PATH', required=True,
+                                 help='path the rotary database directory')
```
jmtsuji: Small grammar edit: should this be changed to `path to the rotary database directory`?

rotary/rotary.py
```diff
@@ -136,6 +139,41 @@ def init(args):
     dataset.create_sample_tsv(output_dir_path, header=sample_tsv_header_fields)
 
 
+def download(args):
+    """
+    Runs code for the download command-line mode. Downloads database files required by rotary.
+
+    :param args: The command-line arguments.
+    """
+    config_path = get_config_path(args, default_config_path=provided_config_path)
```
jmtsuji, line 148: @LeeBergstrand When I reviewed the code, I was a little confused about how the config file is handled. It looks like this function (and the rotary CLI) can accept a user's config file as input. However, this function also writes `download_config.yaml` as output, and it seems like it can generate `download_config.yaml` using the rotary default config if the user doesn't supply a config file. Would it be better to not write `download_config.yaml` if a user-supplied config is input? Or (as an alternative approach) would it be better to just eliminate the option to provide a user-supplied config file? I'm not sure what info the user could  supply in their config file that isn't already known by the download function. Let me know your thoughts or if I am misunderstanding how the code works -- thanks.

rotary/rotary.py
```diff
@@ -136,6 +139,41 @@ def init(args):
...
+
+    output_dir_path = os.getcwd()
+
+    config = setup_database_dir(args, config, output_dir_path)
+
+    jobs = get_cli_arg(args, 'jobs')
+    snakemake_args = get_snakemake_args(args)
+
+    # Tell snakemake to only run the download rule.
+    snakemake_args.append('download')
```
jmtsuji, line 166: I just want to check -- do you also need to add `--until` as an arg before `download` for this to work? Thanks.

rotary/rules/rotary.smk
```diff
@@ -1,21 +1,31 @@
...
+    if SAMPLES.files_per_sample==3:
+        DATASET_HAS_SHORT_READS=True
+    else:
+        DATASET_HAS_SHORT_READS=False
+
+    SAMPLE_NAMES = list(SAMPLES.identifiers)
 else:
+    # Some variables are required for evaluating the DAG because they are used in rules.
+    # We mock them if no samples.tsv file is found.
+    SAMPLE_NAMES=['mock_sample_name']
```
jmtsuji, line 26: Nice that you were able to find a workaround for `samples.tsv` if the download command is run.

jmtsuji (changes requested):
@LeeBergstrand Thanks for working on this enhancement. Looks good to me, pending a couple minor questions and edits. Look forward to your response.

rotary/rotary.py
LeeBergstrand: @jmtsuji In Snakemake, you can add a single rule at the very end of the command, and it will cherrypick out that rule, run only that rule, and then evaluate the DAG based on that rule's required inputs. That's how you run Snakemake if you don't have an `all` rule that gets evaluated automatically. So you could have a collection of rules in a single file and call them individually. If you look at the snakemake CLI page, you will note that at the end of the command, there is an option for `targets` (https://snakemake.readthedocs.io/en/stable/executing/cli.html). By putting the rule 'download' at the end of the command, I'm telling Snakemake to target only that rule.

When you use `--until,` it evaluates the complete DAG and runs Snakemake until it hits that rule. When I used `--until,` the code failed during DAG evaluation due to missing certain files and config parameters; you skipped the full DAG evaluation by telling it to target the `download` rule and evaluate a DAG from there. This causes you to only evaluate the download rules, which require no inputs other than a download directory.

rotary/rotary.py
LeeBergstrand: The command 'run_snakemake_workflow,' included in Pungi, requires a config path. Snakemake uses this path and the file it points to to add config parameters to the Snakemake workflow. Thus, the code always requires a config file to be present. So I make a temporary config file, `download_config.yaml`, use it during the download process, and then delete it afterward. 

```python
run_snakemake_workflow(config_path=config_path, snake_file_path=snake_file_path, output_dir_path=output_dir_path,
                           jobs=jobs, conda_env_dir_path=conda_env_dir, snakemake_custom_args=snakemake_args)
```

LeeBergstrand: I do this to get the download dir path into the Snakemake workflow (as well as several other parameters like hmm URLs, etc.). You may be able to pass these parameters in as CLI parameters to make Snakemake work without a config file, but this would require some refactoring of the run_snakemake workflow inside Pungi.

rotary/rotary.py
LeeBergstrand: Fixed

LeeBergstrand:
> @jmtsuji Tested with dry run be feel free to review.

Tested with a full run. It works. Note that it will create benchmarking and log folders in the CWD.

rotary/rotary.py
jmtsuji: Got it -- thanks! Yes, this makes sense. I remember this snakemake syntax now.

rotary/rotary.py
jmtsuji: Thanks for the additional explanation. That makes sense that a config file needs to be written to disk to run the snakemake workflow. The thing I was wondering is that the CLI for the download module seems to have a flag where the user can provide their own config file. What happens in that case? Does it override the temp config file, for example? Or am I misunderstanding the code somehow? I'll start a 2nd thread where I highlight that code line (252-ish) specifically. Thanks!

rotary/rotary.py
```diff
@@ -192,14 +234,29 @@ def parse_cli():
...
                              help="do not search for short read files")
     parser_init.set_defaults(init=True)
+
+    # =========================
+    # Declare Download Sub-command
+    download_help = """Downloads databases required by rotary to a specific directory """
+    parser_download = subparsers.add_parser('download', help=download_help)
+    parser_download.add_argument('-d', '--database_dir', metavar='PATH', required=True,
+                                 help='path to the rotary database directory')
+    parser_download.add_argument('-c', '--config', metavar='YAML',
```
jmtsuji: @LeeBergstrand Here is the line I was wondering about above (in the [thread about Line ~148](https://github.com/rotary-genomics/rotary/pull/175/files#r1647853209)). It looks like the user can supply their own config file to the CLI. How does that interact with the temporarily generated config file we discussed above?

If the user-provided config file is optional but overrides the temporarily generated one, we might want to mention this in the help statement in the CLI (and make the argument optional). Alternatively, we could just remove this `--config` argument so that the temporarily generated config file is always used -- I wonder if this might prevent confusion for the user. Again, let me know if I am not understanding something. Thanks!

jmtsuji:
> > @jmtsuji Tested with dry run be feel free to review.
> 
> Tested with a full run. It works. Note that it will create benchmarking and log folders in the CWD.

Great that this is tested and working! 👍 Creating the benchmarking and log folders in the CWD seems OK for now. Long-term, we could consider adding a flag to specify the working dir for the download module, but I don't think this is urgent.

rotary/rotary.py
LeeBergstrand: > How does that interact with the temporarily generated config file we discussed above?

@jmtsuji Right now that cli flags takes the specified config, modifies it for jobs, ram, and database directory and writes it as `download_config.yaml`. If the CLI flag is not specified it uses the default prepackaged config.

@jmtsuji I will remove the flag.

rotary/rotary.py
LeeBergstrand: Addressed

LeeBergstrand:
@jmtsuji Decided to remove the CLI config flag. I originally had this so the user could specify what databases to download. However, after reviewing my code my code ignores those config changes and downloads everything anyways. I decided to remove the config cli parameter to make things simpler.

jmtsuji (approved):
@LeeBergstrand Thanks for the additional background. It makes sense that you wanted to leave the config argument so that the user could specify specific databases to download. Only downloading specific DBs might become more important down the road if we ever added support for many different annotation methods (e.g., InterProScan vs. e-mapper vs. ___) and most users only needed to use one method. However, at this stage of development for rotary, I think it should be reasonable to just download all DBs with the download module. (As a workaround, if a user really wanted to only download certain DBs, I suppose they could run `rotary run` directly, without first running the download module.) Overall, sounds good to remove the config argument -- feel free to merge this in!

PR #186, opened 2024-06-05
Snakemake 8

### LINKED ISSUE #111: Move to Snakemake 8.0

LeeBergstrand:
Modify the code to work with snakemake version 8, a significant non-backwards compatible version update.

Addresses issue: https://github.com/rotary-genomics/rotary/issues/111

Being done in coordination with https://github.com/rotary-genomics/pungi/pull/4

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -5,11 +5,11 @@ channels:
 dependencies:
   - python=3.9.*
   - pandas=1.4.*
-  - snakemake=7.*
-  - mamba
-  - wget
-  - psutil=5.9.5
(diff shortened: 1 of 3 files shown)


LeeBergstrand:
@jmtsuji Passes a run of rotary.

Issue #193, opened 2024-06-19
Rotary fails to repolish circular contigs due to workflow errors.


LeeBergstrand:
###  Problem Description

When running rotary on specific FASTQ files, I get the following error:

```
WorkflowError in rule stitch_medaka in file /home/lee_bergstrand/rnd/rotary/rotary/rules/./polish.smk, line 127:
Failed to open input file: STR32-b/circularize/medaka/STR32-b_contigs.txt. Has it been deleted by another process? (rule stitch_medaka, line 221, /home/lee_bergstrand/rnd/rotary/rotary/rules/./polish.smk)
```

This rule occurs after the DAG is re-evaluated following the triggering of the checkpoint `split_circular_and_linear_contigs`

I noticed that the directories `medaka` and `medaka_input` do not exist under the `circularize` directory. The only directories under `circularize` are `filter/lists` with the linear and circular contig lists. No FASTA files were generated under `circularize`.

LeeBergstrand:
The above error occurs when running snakemake 8.0 (develop_lee branch). When running snakemake 7.0 (develop branch), I get the following error message:

```
BUG: Out of jobs ready to be started, but not all files built yet. Please check https://github.com/snakemake/snakemake/issues/823 for more information.
Remaining jobs:
 - combine_circular_and_linear_contigs: SO11/circularize/combine/SO11_combined.fasta
 - finalize_circular_contig_rotation: SO11/circularize/combine/SO11_circular.fasta
 - generate_contig_manifest: SO11/circularize/medaka/SO11_contigs.txt
 - prepare_medaka_circularize_input: SO11/circularize/medaka_input/SO11_input.fasta
 - all: 
 - run_circlator: SO11/circularize/circlator/SO11_rotated.fasta, SO11/circularize/circlator, SO11/circularize/circlator/rotated.promer.contigs_with_ends.fa
 - get_polished_contigs: SO11/circularize/filter/SO11_circular.fasta
 - get_start_genes: SO11/circularize/identify/SO11_start_gene.ffn
 - search_contig_start: SO11/circularize/identify/SO11_circular.faa, SO11/circularize/identify/SO11_circular.ffn, SO11/circularize/identify/SO11_circular.gff, SO11/circularize/identify/SO11_hmmsearch_hits.txt, SO11/circularize/identify/SO11_hmmsearch_hits_no_comments.txt
 - process_start_genes: SO11/circularize/identify/SO11_start_genes.list
[...]

PR #194, opened 2024-06-21
Fix circularize medaka missing contigs file

LeeBergstrand:
Potentially fixes issue https://github.com/rotary-genomics/rotary/issues/193


LeeBergstrand:
@jmtsuji When is medaka ran during the circularization process? Is it only ran on the circular contigs? Since the circular contig list only contains the circular contigs names it is identical to the contig manifest that medaka needs. So I take a copy and place that where the snakemake code is looking.

PR #200, opened 2024-07-02
raise value error if no contigs pass coverage filter.

### LINKED ISSUE #199: Rotary Hangs When No Contigs Pass Coverage Filter
### Problem Description

In certain situations with lousy input data, Rotary can determine that no contigs pass the coverage filter, causing checkpoint' split_circular_and_linear_contigs' to write no circular or linear contig lists. This, in turn, causes rule `combine_circular_and_linear_contigs` to fail because the function `aggregate_contigs` returns an empty input list. This, in turn, causes the command `cat | seqtk seq -l 60 > {output}` to run, as `{input}` is empty. This command has infinite run time. 

### Proposed Solution

Throw a Python exception when no contigs pass the coverage filter.

LeeBergstrand:
Resolves issue https://github.com/rotary-genomics/rotary/issues/199 -- An error occurs when no contigs pass the coverage filter.

### DIFF
--- a/rotary/rules/circularize.smk
+++ b/rotary/rules/circularize.smk
@@ -47,6 +47,9 @@ checkpoint split_circular_and_linear_contigs:
         contig_info = pd.read_csv(input[0], sep='\t')
         contig_info_filtered = contig_info[contig_info['pass_coverage_filter'] == 'Y']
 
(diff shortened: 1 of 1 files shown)


LeeBergstrand:
@jmtsuji Does this solution work for you?

PR #214, opened 2024-07-19
Merge develop_lee into develop

LeeBergstrand:
I have merged all the current backlog of pull requests into develop_lee as they are piling up.

Features Added:

- Raise a value error when no contigs pass the coverage test in rule `split_circular_and_linear_contigs.` Fixes https://github.com/rotary-genomics/rotary/issues/199 where no circular or linear contigs lists are written, which causes a workflow error.
- Update the QC before-and-after report code to handle KBP (highly filtered reads) and GBP (many Illumina reads) values. These values commonly break the old version of the code. Convert results with these units to MBP - See https://github.com/rotary-genomics/rotary/pull/201 for details.
- In the rule `search_contig_start`, add the `-p meta` flag to run prodigal in meta mode. The prodigal doesn't scan the genome in this mode to build the gene-calling model. Instead, it predicts genes using prebuilt models, which is suitable for metagenomics. Prodigal requires at least 20K bases to make a model, which is larger than many circular plasmids. In some cases, Prodigal would fail when the only circular element is a small plasmid because the rest of the genome is fragmented and the plasmid is shorter than 20K bases. @jmtsuji Do you think running the `-p meta` flag should be sufficient if we are doing a start gene search? 
- Remove end_repair assemblies from assembly stats as they are most often identical to the flye assembly stats.
- Aggregate checkm stats from all samples into a single file for easier viewing.
- Aggregate gtdbtk reports from all samples into a single file for easier viewing.
- Aggregate coverage report from all samples into a single file for easier viewing.
- Change the top-level stats folder name to 'aggregate_stats' to indicate that this folder contains stats from multiple samples
- Add freebayes explicitly to the `pyploca.yml` install to prevent an install failure.
- Snakemake 8.6.0 introduced a bug that prevents the deletion of temp files created before a checkpoint, causing storage issues due to accumulating gigabytes of temp files, which I reported in snakemake/snakemake#2982. Downgrade to version 7 to temporarily fix this issue. Details: https://github.com/rotary-genomics/rotary/pull/216. I was originally going to go back to snakemake version 8.5.5 but I found that this version also causes some stability issues (I think it was before version 8's initial release), so I decided to downgrade to version 7. Details: https://github.com/rotary-genomics/rotary/pull/217
- Modify rule `set_up_sample_directories` so that each sample directory is generated dynamically and remove the sample tsv as input. This allows the sample TSV to be modified without causing the entire pipeline to be run all over again from scratch.
- Fixes a minor bug where aggregated CheckM2 and GTDB-Tk could not be turned off in the config file, because they were not specified as optional in rule annotation in annotations.smk. They are now specified as optional,

### DIFF
--- a/rotary/rules/assembly.smk
+++ b/rotary/rules/assembly.smk
@@ -92,8 +92,7 @@ rule assembly_end_repair:
 
 rule run_quast:
     input:
-        flye_assembly = "{sample}/assembly/flye/{sample}_flye_assembly.fasta",
-        end_repair_assembly= "{sample}/assembly/end_repair/{sample}_end_repair_assembly.fasta",
+        flye_assembly = "{sample}/assembly/flye/{sample}_flye_assembly.fasta"
     output:
         stats_dir=directory("{sample}/assembly/stats/"),
         report=multiext("{sample}/assembly/stats/report", '.html', '.tex', '.tsv', '.txt', '.pdf'),
--- a/rotary/rules/assembly.smk
+++ b/rotary/rules/assembly.smk
@@ -108,7 +108,7 @@ rule run_quast:
     shell:
         """
         quast -t {threads} --space-efficient \
-        --labels "{wildcards.sample}_flye, {wildcards.sample}_end_repair" \
+        --labels "{wildcards.sample}_flye" \
         -o {output.stats_dir} {input} > {log} 2>&1
         """
 
--- a/rotary/rules/assembly.smk
+++ b/rotary/rules/assembly.smk
@@ -133,8 +133,8 @@ rule aggregate_assembly_stats_multiqc:
     input:
         expand("{sample}/assembly/stats/",sample=SAMPLE_NAMES)
     output:
-        multqc_report="stats/assembly/all_samples_assembly_stats_multiqc_report.html",
-        multqc_report_data="stats/assembly/all_samples_assembly_stats_multiqc_report_data.zip"
+        multqc_report='aggregate_stats/assembly/all_samples_assembly_stats_multiqc_report.html",
+        multqc_report_data='aggregate_stats/assembly/all_samples_assembly_stats_multiqc_report_data.zip"
     conda:
         "../envs/qc.yaml"
     log:
@@ -142,7 +142,7 @@ rule aggregate_assembly_stats_multiqc:
     benchmark:
         "benchmarks/assembly/assembly_stats_multiqc.txt"
     params:
-        qc_stats_dir="stats/assembly/"
+        qc_stats_dir='aggregate_stats/assembly/"
     shell:
         """
         multiqc --outdir {params.qc_stats_dir} \
@@ -156,7 +156,7 @@ rule assembly:
     input:
         assembly_fasta = expand("{sample}/assembly/{sample}_assembly.fasta",sample=SAMPLE_NAMES),
         assembly_circular_info = expand("{sample}/assembly/{sample}_circular_info.tsv", sample=SAMPLE_NAMES),
-        assembly_multqc_report="stats/assembly/all_samples_assembly_stats_multiqc_report.html",
-        assembly_multqc_report_data="stats/assembly/all_samples_assembly_stats_multiqc_report_data.zip"
+        assembly_multqc_report='aggregate_stats/assembly/all_samples_assembly_stats_multiqc_report.html",
+        assembly_multqc_report_data='aggregate_stats/assembly/all_samples_assembly_stats_multiqc_report_data.zip"
     output:
         temp(touch("checkpoints/assembly"))
--- a/rotary/rules/assembly.smk
+++ b/rotary/rules/assembly.smk
@@ -108,7 +108,7 @@ rule run_quast:
     shell:
         """
         quast -t {threads} --space-efficient \
-        --labels "{wildcards.sample}_flye" \
+        --labels "{wildcards.sample}" \
         -o {output.stats_dir} {input} > {log} 2>&1
         """
 
--- a/rotary/rules/assembly.smk
+++ b/rotary/rules/assembly.smk
@@ -133,8 +133,8 @@ rule aggregate_assembly_stats_multiqc:
     input:
         expand("{sample}/assembly/stats/",sample=SAMPLE_NAMES)
     output:
-        multqc_report='aggregate_stats/assembly/all_samples_assembly_stats_multiqc_report.html",
-        multqc_report_data='aggregate_stats/assembly/all_samples_assembly_stats_multiqc_report_data.zip"
+        multqc_report="aggregate_stats/assembly/all_samples_assembly_stats_multiqc_report.html",
+        multqc_report_data="aggregate_stats/assembly/all_samples_assembly_stats_multiqc_report_data.zip"
     conda:
         "../envs/qc.yaml"
     log:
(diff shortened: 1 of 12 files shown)


rotary/rules/assembly.smk
```diff
@@ -92,8 +92,7 @@ rule assembly_end_repair:
 
 rule run_quast:
     input:
-        flye_assembly = "{sample}/assembly/flye/{sample}_flye_assembly.fasta",
-        end_repair_assembly= "{sample}/assembly/end_repair/{sample}_end_repair_assembly.fasta",
+        flye_assembly = "{sample}/assembly/flye/{sample}_flye_assembly.fasta"
```
jmtsuji, line 95: @LeeBergstrand During the polish phase, low coverage contigs are filtered out of the assembly. Because of this, I think it might make sense to add another input line for the post-polish genome, e.g.,

```
polish_assembly= "{sample}/polish/{sample}_polish.fasta"
```

What do you think?

jmtsuji: If we make this change, then this QUAST rule might work better in a different snakemake file than `assembly.smk`. For example, we might need to make a separate snakefile like `aggregate.smk`...

rotary/rules/circularize.smk
```diff
@@ -47,6 +47,9 @@ checkpoint split_circular_and_linear_contigs:
         contig_info = pd.read_csv(input[0], sep='\t')
         contig_info_filtered = contig_info[contig_info['pass_coverage_filter'] == 'Y']
 
+        if len(contig_info_filtered) == 0:
```
jmtsuji, line 50: Helpful addition -- thanks.

rotary/rules/circularize.smk
```diff
@@ -101,7 +104,7 @@ rule search_contig_start:
     shell:
         """
         printf "### Predict genes ###\n" > {log}
-        prodigal -i {input.contigs} -a {output.orf_predictions} -d {output.gene_predictions} \
+        prodigal -i {input.contigs} -a {output.orf_predictions} -d {output.gene_predictions} -p meta \
```
jmtsuji, line 107: Agreed, let's use `-p meta` for these rough ORF predictions.

rotary/rules/qc.smk
```diff
@@ -495,8 +495,8 @@ rule generate_short_aggregate_qc_stats:
     input:
         expand("{sample}/qc/qc_stats/short/{sample}_short_multiqc_report_data.zip", sample=SAMPLE_NAMES)
     output:
-        left_qc_stats='stats/qc/before_and_after_qc_sequence_stats_short_R1.tsv',
-        right_qc_stats='stats/qc/before_and_after_qc_sequence_stats_short_R2.tsv'
+        left_qc_stats='aggregate_stats/qc/before_and_after_qc_sequence_stats_short_R1.tsv',
```
jmtsuji, line 498: @LeeBergstrand These aggregated summary files are very helpful -- thanks! What might you think about combining the R1 and R2 short read stats into a single file, for conciseness? Another column of R1 vs. R2 could be added to each existing file, and then the R1 and R2 files could be concatenated vertically.

jmtsuji: This could be done in a future PR so we can go ahead and merge this PR.

rotary/rules/circularize.smk
jmtsuji: We do not need to change anything for this PR, but one other idea I've played around with is using DFAST (without functional annotations, just gene calling) for this step, to ensure that the called genes during this phase are consistent with those in the final assembly. Any thoughts on this idea?

jmtsuji (changes requested):
@LeeBergstrand thanks so much for making this PR and for adding a concise summary of changes. Overall, these looks like a nice set of changes for adding stability and enhancing reporting. The complete end-to-end test on my end (for one sample) passed flawlessly.

I've added a couple very minor comments in my code review that can be quickly addressed, along with a couple brief discussions point. I also reply to a few of your specific comments in the PR description below.

Really appreciate your work on this, and sorry for taking a long time to add a review. Other than the discussion points I left, this should be good to merge.

## Comments on your PR description

I am OK with most of the comments you left in your PR description, so I will just reply to a few of them below.

> In the rule search_contig_start, add the -p meta flag to run prodigal in meta mode. The prodigal doesn't scan the genome in this mode to build the gene-calling model. Instead, it predicts genes using prebuilt models, which is suitable for metagenomics. Prodigal requires at least 20K bases to make a model, which is larger than many circular plasmids. In some cases, Prodigal would fail when the only circular element is a small plasmid because the rest of the genome is fragmented and the plasmid is shorter than 20K bases. @jmtsuji Do you think running the -p meta flag should be sufficient if we are doing a start gene search?

I comment on this in the code review, but agreed -- let's change to -p to support small assemblies.

> Remove end_repair assemblies from assembly stats as they are most often identical to the flye assembly stats.

OK -- sounds reasonable. You'll see in the code review that I suggest to add the post-polishing assembly to the assembly stats because some contigs can be removed during length filtration.


> Snakemake 8.6.0 introduced a bug that prevents the deletion of temp files created before a checkpoint, causing storage issues due to accumulating gigabytes of temp files, which I reported in Temp files are not deleted for rule before checkpoint that is rerun after checkpoint with new data.  snakemake/snakemake#2982. Downgrade to version 7 to temporarily fix this issue. Details: Fix snakemake version to be less than 8.6.0 to fix temp delete issue. #216. I was originally going to go back to snakemake version 8.5.5 but I found that this version also causes some stability issues (I think it was before version 8's initial release), so I decided to downgrade to version 7. Details: Move rotary back to using snakemake 7 due to issues with snakemake 8 #217

Thanks for working on this. Too bad about snakemake 8... I guess we need to wait for now until an update is released. Thanks for making an issue in the snakemake repo about the issue, too.

> Aggregate checkm stats from all samples into a single file for easier viewing.
> Aggregate gtdbtk reports from all samples into a single file for easier viewing.
> Change the top-level stats folder name to 'aggregate_stats' to indicate that this folder contains stats from multiple samples

This is a great addition - thanks. In addition to stats, I wonder if it would make sense to package all of the final genomes into a ZIP file or something that could accompany the summary stats -- what do you think? This could replace (or supplement) the current per-genome ZIP folder.

> Modify rule set_up_sample_directories so that each sample directory is generated dynamically and remove the sample tsv as input. This allows the sample TSV to be modified without causing the entire pipeline to be run all over again from scratch.

Cool change!


Look forward to your feedback. Thanks again.

rotary/rules/circularize.smk
LeeBergstrand: @jmtsuji I think using dfast for this step could work. I want to discuss with you about dfast vs prodigal in our next meeting.

rotary/rules/qc.smk
LeeBergstrand: @jmtsuji I open to different formats. That makes sense.

rotary/rules/circularize.smk
jmtsuji: @LeeBergstrand Sounds good. I'm not sure how well DFAST would perform on a highly fragmented input genome... let's discuss more later.

rotary/rules/qc.smk
jmtsuji: OK -- I'll convert this to an issue so we can consider it later.

jmtsuji: Now in #222

rotary/rules/circularize.smk
jmtsuji: Update: not sure that DFAST could handle annotating just a short circular plasmid properly. Let's just keep on using prodigal.

rotary/rules/assembly.smk
jmtsuji: Converted to issue #224

jmtsuji:
@LeeBergstrand All comments are now addressed. It might make sense to first review and merge #223 into this branch before merging into develop, but otherwise, #223 could be handled as a separate PR directly into develop. So feel free to merge this PR (#214) when ready.

PR #216, opened 2024-07-23
Fix snakemake version to be less than 8.6.0 to fix temp delete issue.

LeeBergstrand:
Snakemake 8.6.0 introduced a bug that prevented the deletion of temp files created before a checkpoint, even after a complete DAG execution. This bug was introduced when a pull request was merged into version 8.6.0 of Snakemake, which prevented some corner cases where temp files were deleted before a checkpoint even though they were needed later. However, this change prevents any temp files before a checkpoint from being deleted. After this bug was introduced, these temp files were not deleted, even at the end of the pipeline. We have many gigabytes of temp files in our pipeline before the checkpoint 'split_circular_and_linear_contigs' that need to be deleted; this leads to my server running out of storage room.  I provided a bug report to Snakemake here: https://github.com/snakemake/snakemake/issues/2982

As a temporary fix I am fixing the Snakemake version to be less than 8.6.0 but more than 8.0.0.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -5,7 +5,7 @@ channels:
 dependencies:
   - python=3.11.*
   - pandas=1.5.*
-  - snakemake=8.*
+  - snakemake>=8,<8.6.0
   - mamba=1.5.*
   - wget=1.21.*
   - psutil=5.9.*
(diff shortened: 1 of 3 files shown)


LeeBergstrand:
I figured out that I basically have to go back farther to version 7.0 because older versions of snakemake 8 are not stable and cause issues.

